PR #11, opened 2016-01-12
Update setup and test scripts

darrenleeweber:


lib/moab/utc_time.rb
```diff
@@ -0,0 +1,39 @@
+require 'time'
+
+module Moab
+  # Timestamp conversion methods.
+  class Moab::UtcTime
+
+    # @param datetime [Time,String,Nil] The input datetime
+    # @return [void] Convert input datetime to a Time object, or nil if input is empty.
```
atz: return is *not* void!

cbeer: This class is just moved from the monkey patches (and.. isn't a monkey patch)

atz:
:+1: to fixing bad habits.

darrenleeweber:
This was merged with potentially significant namespace changes that can impact down-stream projects.  It's not clear that these downstream projects are version pinned to cope with these changes.  All of this should have been done in a feature branch off of develop and, as one of the DLSS devs with more experience on this project, I should be consulted on it before merging.

PR #30, opened 2017-09-01
spec/unit_tests/stanford: remove cruft (inspired by rubocop)

ndushay:
Please review especially for:
- no useful tests removed  (some pointless ones have been removed)
- no existing test functionality has been changed (refactored some variable names or avoided long lines or changed otherwise for readability)

This spec cleanup:

- shortens names of specs to be just the method names
- removes comments containing the rdoc or implementation of methods
- removes specs that are useless  (e.g. that setting an attribute value means you can retrieve the attribute value;  that instantiating an object returns an instance of the object)
- splits out multiple tests in a single spec into separate tests
- avoids long lines
- refactors variable names for clarity
- removes unnecessary @ variables
- tried to follow principles to make rubocop happier
etc.

### DIFF
--- a/spec/unit_tests/stanford/active_fedora_object_spec.rb
+++ b/spec/unit_tests/stanford/active_fedora_object_spec.rb
@@ -1,79 +1,23 @@
 require 'spec_helper'
 
-# Unit tests for class {Stanford::ActiveFedoraObject}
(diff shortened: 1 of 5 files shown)


ndushay:
I just changed the coveralls threshold for decrease in coverage to 0.1%;  will hopefully avoid false errors in future.

PR #34, opened 2017-09-01
Add object size method to gem

### LINKED ISSUE #21: moab-vers: add method to get object size (for online, exploded objects) and update README
Since we'll probably want to be able to do this from various places, it would be useful for this code to live in the gem.

SaravShah:
- Add object size method to `Moab::StorageRepository`
- Updated README w/ usage

closes #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ current_version = StorageServices.current_version('666') # where 666 is the id
 expect(current_version).to be_an_instance_of Integer
 ```
 
+#### Get Size of Moab Object
+```ruby
+object_size_in_bytes = StorageServices.object_size('666') # where 666 is the id
+expect(current_version).to be_an_instance_of Integer
+```
+
 ### Stanford-Specific
 
 #### Configuration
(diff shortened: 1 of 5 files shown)


ndushay:
I agree that it looks good and that adding the test he suggests would be good.

Nice job!

README.md
```diff
@@ -56,6 +56,13 @@ current_version = Stanford::StorageServices.current_version('oo000oo0000') # whe
 expect(current_version).to be_an_instance_of Integer
 ```
 
+#### Get Size of Moab Object
+
+```ruby
+object_size_in_bytes = Stanford::StorageServices.object_size('oo000oo0000') # where oo000oo0000 is the druid
+expect(object_size_in_bytes).to be_an_instance_of Integer
```
ndushay: I think the implementation isn't stanford specific, so perhaps the README changes should be in both generic and stanford specific places??

SaravShah: agreed. just updated it!

PR #35, opened 2017-09-02
[WIP] debugging tests

jmartin-sul:
a couple possible approaches to getting the tests to pass for #34.  also, a possible way to slightly reduce duplicated code.  i think the second approach (a mock result for `Moab::StorageRepository#storage_branch`) might be the better approach, but i'm not sure.  the second approach does seem more in line with the rest of that test file, and the `#storage_branch` method is already tested elsewhere (and is meant to be overridden anyway, according to the comments).

i think the way to go for actually fixing things isn't to merge this into the branch that this PR is against, but to just reimplement the changes on that branch and squash the result of that back down to 1 commit.  or, pull this into that branch, remove the remaining debugging statement, do whatever other cleanup, and squash it back down to 1 commit.  whichever's easier.  not much code should end up getting changed here, and it doesn't feel like more than a single commit.

### DIFF
--- a/lib/moab/storage_repository.rb
+++ b/lib/moab/storage_repository.rb
@@ -102,6 +102,23 @@ module Moab
       storage_object
     end
 
+    def object_size(object_id)
+      root = find_storage_root(object_id)
+      p root
+      storage_pathname = root.join(storage_trunk,storage_branch(object_id)).to_s
+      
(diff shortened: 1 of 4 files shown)


SaravShah:
yeah i think re-implementing will be easier for me!

PR #37, opened 2017-09-05
Add documentation for how to make a Moab::FileInventory

### LINKED ISSUE #22: moab-vers:  be able to retrieve expected checksums for a version(?)
- figure this out
- document how to do this in the moab-versioning README
- are the moab existing checksums  in `signatureCatalog.xml` ??

investigation:
- there is probably already code to do this -- look at specs?  look at code? 
- look for code that reads this info from moab
  - if none, figure it out?  add it?

from StoryTime(tm):
>We will need to parse XML files from Moab object in order to get out checksum information. Perhaps the moab object traversal and pulling checksums out of its xml files should be part of moab gem.

tallenaz:
Closes #22

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,41 @@ object_size_in_bytes = Stanford::StorageServices.object_size('oo000oo0000') # wh
 expect(object_size_in_bytes).to be_an_instance_of Integer
 ```
 
+#### Get Inventory From Content Metadata
+
+To generate a `Moab::FileInventory` object containing fixity, size, and other info:
+
+```ruby
+require 'moab/stanford'
+doc = IO.read('path/to/contentMetadata.xml')
+sci = Stanford::ContentInventory.new
+sci.inventory_from_cm(doc, 'druid:th154ru1456', 'all', '10') # all of v0010
+=> #<Moab::FileInventory:0x007fdcd9435888
+ @digital_object_id="druid:th154ru1456",
+ @groups=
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,6 +56,17 @@ current_version = Stanford::StorageServices.current_version('oo000oo0000') # whe
...
+
+To generate a `Moab::FileInventory` object containing fixity, size, and other info:
+
+```ruby
+require 'moab/stanford'
+doc = IO.read('path/to/contentMetadata.xml')
+sci = Stanford::ContentInventory.new
+sci.inventory_from_cm(doc,'druid:th154ru1456','all','10') # all of v0010
+```
+
```
ndushay: 3 things:

1.  is this stanford specific?  If not, move it above line 37

2.  it might be nice to show a snippet of example output, or some rspec-y stuff to show what the result will be

3.  spaces after commas, please

jmartin-sul: agreed, and to add one more nitpick, i'd say that the `require 'moab/stanford'` seem unnecessary, just because it's mentioned first thing under "Usage", and because the other examples don't include it.

tallenaz:
@ndushay and @jmartin-sul re: your requested changes:

I believe this is Stanford specific, because I can only invoke the method from a `Stanford::ContentInventory` object, and can only generate one of those by specifically requiring `moab/stanford`

Otherwise, I've added sample output (hopefully not too lengthy) with fake druid and checksum info, and spaced out commas.

PR #38, opened 2017-09-06
Git ignore .ruby-gemset file

tingulfsen:


jmartin-sul:
rebased off current master and re-pushed, to get rid of rubocop error that was fixed in a prior merged PR

jmartin-sul:
merging without waiting for coveralls, which seems to be stalled in reporting, because this shouldn't change test coverage

PR #46, opened 2017-09-10
unit_tests/moab/file_inventory_difference_spec: remove cruft

ndushay:
Please review especially for:
- no useful tests removed  (some pointless ones have been removed)
- no existing test functionality has been changed (refactored some variable names or avoided long lines or changed otherwise for readability)

This spec cleanup:

- shortens names of specs to be just the method names
- removes comments containing the rdoc or implementation of methods
- removes specs that are useless  (e.g. that setting an attribute value means you can retrieve the attribute value;  that instantiating an object returns an instance of the object)
- splits out multiple tests in a single spec into separate tests
- avoids long lines
- refactors variable names for clarity
- removes unnecessary @ variables
- tried to follow principles to make rubocop happier
etc.

### DIFF
--- a/spec/unit_tests/moab/file_inventory_difference_spec.rb
+++ b/spec/unit_tests/moab/file_inventory_difference_spec.rb
@@ -25,81 +25,43 @@ describe 'Moab::FileInventoryDifference' do
     end
   end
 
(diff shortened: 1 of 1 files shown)


SaravShah:
this looks good to me 👍

PR #51, opened 2017-09-13
fixed current_version to object_size_in_bytes

SaravShah:
changed 1 variable name that was there from copy pasting


jmartin-sul:
merged before coveralls reported, because tests pass and a readme change shouldn't affect code coverage

PR #58, opened 2017-10-13
Size fix

### LINKED ISSUE #56: add more efficient method to get object_size if storage_dir is known
e.g., for version, this is more efficient

```ruby
moab = Moab::StorageObject.new(druid, object_dir) # cheaper/faster to go directly to the correct directory
current_version = moab.current_version_id 
```

than

```ruby
current_version = Stanford::StorageServices.current_version(druid)
```

because the latter looks through each storage_root until it finds the object.

ndushay:
adds a more efficient way to get object size information, if you already know the storage root.

Also threw in a teensy bit of cleanup.

Fixes #56


lib/moab/file_group_difference.rb
```diff
@@ -263,14 +263,13 @@ def tabulate_renamed_files(matching_signatures, basis_signature_hash, other_sign
...
-            else
-              fid.change = 'renamed'
+          if fid.basis_path.nil?
+            fid.change = 'copyadded'
+            fid.basis_path = basis_paths[0]
+          elsif fid.other_path.nil?
+            fid.change = 'copydeleted'
+          else
+            fid.change = 'renamed'
           end
```
darrenleeweber: Wondering whether the `case` is used here to explicitly allow it to fall-through from one `when` to the next and so on; a cascading flow of checking/correcting values for `fid.change`?

lib/moab/storage_object.rb
```diff
@@ -153,11 +153,7 @@ def empty?
     # @api external
     # @return [Integer] The identifier of the latest version of this object, or 0 if no versions exist
     def current_version_id
-      return @current_version_id unless @current_version_id.nil?
-      #@current_version_id = self.version_id_list.last || 0
-      list = self.version_id_list
-      version_id = list.empty? ? 0 : list.last
-      @current_version_id = version_id
+      @current_version_id ||= self.version_id_list.last || 0
     end
```
darrenleeweber: It's curious that the old code had commented out the `@current_version_id = self.version_id_list.last || 0`, as if that was considered and rejected for some reason (some bug).

lib/moab/storage_object.rb
```diff
@@ -230,6 +226,18 @@ def restore_object(recovery_path)
...
+      size = 0
+      Find.find(object_pathname) do |path|
+        if FileTest.directory?(path)
+          Find.prune if File.basename(path)[0] == '.'
+        else
+          size += FileTest.size(path)
+        end
+      end
+      size
+    end
```
darrenleeweber: Also prune `..` directory?

lib/moab/file_group_difference.rb
jmartin-sul: ruby doesn't do fall-through on case statements:
https://docs.ruby-lang.org/en/2.4.0/syntax/control_expressions_rdoc.html#label-case+Expression
https://stackoverflow.com/questions/31443922/does-rubys-case-statement-fall-through
https://ruby-doc.org/docs/keywords/1.9/Object.html#method-i-case
https://www.ruby-forum.com/topic/98827

(docs don't say it explicitly, but they do imply a lack of fall-through, and the internet comments seem to confirm this)

lib/moab/storage_object.rb
jmartin-sul: AFAICT from a quick test on the rails console, neither of the implicit `.` or `..` dirs appears in the list of dirs that `Find.find` iterates over.

lib/moab/storage_object.rb
```diff
@@ -231,6 +227,18 @@ def restore_object(recovery_path)
...
 
+    # @return [Integer] the size occupied on disk by the storage object, in bytes.  this is the entire moab (all versions).
+    def size
+      size = 0
+      Find.find(object_pathname) do |path|
+        if FileTest.directory?(path)
+          Find.prune if File.basename(path)[0] == '.'
+        else
+          size += FileTest.size(path)
+        end
```
jmartin-sul: it seems that neither `.` nor `..` are included in the results (which makes sense, because then failing to prune them proactively would result in infinite loop behavior by default)... as such, what about:
```ruby
size += FileTest.size(path) unless FileTest.directory?(path) # must aggregate size across all individual files
```

instead of the explicit `if/else` above?

e.g., if i do the following on a directory called `spec/fixtures/empty`, which only contains one file, called `.keep`, i get:
```ruby
[1] pry(main)> Find.find('spec/fixtures/empty').map { |dir| dir }
=> ["spec/fixtures/empty", "spec/fixtures/empty/.keep"]
[2] pry(main)> 
```

which implies to me that pruning the implicit `.` and `..` is unnecessary.

also, i don't feel strongly at all about the comment i added above, i just thought it might be nice to explain that we can't call `size` directly on a directory.  though that's sort of implied anyway if you assume we're somewhat sensible and not writing extra code when we could've just called a `size` method on the top level dir.

lib/moab/storage_object.rb
jmartin-sul: seems like this should work fine whenever the code being replaced would've worked.  if `version_id_list` is empty, `#last` will return `nil`, and `@current_version_id` will be set to `0`.  this just collapses what used to be lines 159 and 160 into one `||` statement, as well as collapsing the memoization of 157 and 161 into one `||=` statement.

unless there's some specific worry that `version_id_list` won't be a valid list, i think this is fine.

darrenleeweber: The only problem arises if/when `self.version_id_list == nil`, then `.first` will blow up.  (I dunno why I had some spare time to look at this, but it popped up in my github notifications and I had a bit of time to look today; apologies if I'm poking in where I shouldn't).

ndushay: list.empty? will also throw error if list is nil, so no real change.

darrenleeweber: It looks like `list.empty?` would blow up anyway if the `self.version_id_list` ever returned `nil`.

jmartin-sul: also, `self.version_id_list` should always return a list:
https://github.com/sul-dlss/moab-versioning/blob/b2770c71d3e57515d68976ceb7aa749f81ea8e7a/lib/moab/storage_object.rb#L131-L141
```ruby
    def version_id_list
      list = Array.new
      return list unless @object_pathname.exist?
      @object_pathname.children.each do |dirname|
        vnum = dirname.basename.to_s
        if vnum =~ /^v(\d+)$/
          list << vnum[1..-1].to_i
        end
      end
      list.sort
    end
```

PR #60, opened 2017-10-17
code coverage no longer includes spec files

ndushay:
- spec_helper:  
    - don't report coverage in spec files
    - only report coverage once per travis build (not for each ruby version)
    - cleaned up some cruft not needed or no longer needed
- .travis.yml:  added ruby 2.4.2;  added explicit script so we can run `bundle exec coverall push` if we want to check coverage for a build outside of travis

Note that coverage decrease is because we no longer count spec files.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,11 +1,14 @@
 #$LOAD_PATH.unshift(File.dirname(__FILE__))
 #$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', 'lib'))
 require 'coveralls'
-Coveralls.wear!
+Coveralls.wear_merged! # because we run travis on multiple rubies
 
-require 'rubygems'
-require 'rspec'
-require 'rspec/core'
+require 'simplecov'
+SimpleCov.formatter = Coveralls::SimpleCov::Formatter
+SimpleCov.start do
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -1,26 +1,21 @@
...
+
+require 'simplecov'
+SimpleCov.formatter = Coveralls::SimpleCov::Formatter
+SimpleCov.start do
+  add_filter 'spec'
+end
 
-require 'rubygems'
-require 'rspec'
-require 'rspec/core'
```
ndushay: there is no need for these requires as they are automatic with rspec

spec/spec_helper.rb
```diff
@@ -1,26 +1,21 @@
...
 RSpec.configure do |config|
   config.before(:all) {fixture_setup}
-  # Use Gherkin DSL syntax in specs for better readability
-  # scenario "<title>" do
-  #   given: <inputs>
-  #   action: <the application does>
-  #   outcome: <to generate this result>
-  #   ...
-  # end
-  config.alias_example_to :scenario
```
ndushay: there are no more scenarios because i don't think anyone uses the api docs or keeps them up to date and this syntax was just confusing.   I believe the true documentation is in the working specs and in the code itself.

PR #64, opened 2017-10-19
Refactoring typos and rubocop identified lint, etc.

ndushay:
This PR started with me not being able to stand seeing the typos highlighted in my editor.

Then I got a little carried away -- but I was mostly targeting rubocop complaints that were either addressing potentially confusing code, or simply really easy to fix;  mostly I was only concerned with files in the lib directory.

I don't find it too bad to review the whole thing side by side;  the only "difficult" one is where I removed a useless line of code (assignment to a boolean that is never used) in lib/moab/file_group.rb.


lib/moab/file_group.rb
```diff
@@ -222,9 +222,8 @@ def harvest_directory(path, recursive, validated=nil)
     # @param validated [Boolean] if true, path is verified to be descendant of (#base_directory)
```
jmartin-sul: might want to change the param name here too and say something about it not actually being used in this method?  is the param there just to maintain compatibility with callers that call similar methods and always pass that param in?  it does appear from a cursory grep of the codebase like at least one caller expects to pass that param.

lib/moab/storage_object_version.rb
```diff
@@ -85,7 +85,7 @@ def find_filepath(file_category, file_id)
   # @param [String] file_category The category of file ('content', 'metadata', or 'manifest')
```
jmartin-sul: might want to add an underscore to the param name here too?

ndushay:
@jmartin-sul thanks for the feedback and good 👀 on the rdoc/yard comments.  I believe I have addressed your concerns. and this is ready to go (after travis builds, etc.)

PR #72, opened 2017-10-28
(MV) StorageObjectValidator Class

SaravShah:
TODOs:
- Figure out which methods we want to be public vs private.
- Refactoring suggestions
- will create a ticket for `[ ] content and metadata directories each contain only files`


lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,204 @@
...
+        # and good, then no error message will be added to the result array, so if it stays the same 
+        # size then it is ready to be checked for the data_sub_dirs and manfist_xml
+        before_result_size = results.size
+        results.concat check_sub_dirs(version_sub_dirs, version_dir, EXPECTED_VERSION_SUB_DIRS)
+        after_result_size = results.size
+        # don't bother checking the data sub dir if we didn't find the expected two subdirs
+
+        # Before it was result.size === 0, however for each version that is not correctly constructed, 
+        # error messages will be concat to the array. By the time the loop reaches a version that is 
+        # correctly constructe the size will stay the same as before (because it has the old error messages)
```
tallenaz: typo

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+    def correctly_formed_moab
+      results = []
+
+      version_directories = sub_dirs(storage_obj_path)
+      version_directories.each do |version_dir|
+        version_path = "#{storage_obj_path}/#{version_dir}"
+        version_sub_dirs = sub_dirs(version_path)
+        # Need to add before_result_size and after_result_size because if everything is expected
+        # and good, then no error message will be added to the result array, so if it stays the same
+        # size then it is ready to be checked for the data_sub_dirs and manfist_xml
```
ndushay: I would say this comment adds little to no value ... the code shows it.

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+        before_result_size = results.size
+        results.concat check_sub_dirs(version_sub_dirs, version_dir, EXPECTED_VERSION_SUB_DIRS)
+        after_result_size = results.size
+        # don't bother checking the data sub dir if we didn't find the expected two subdirs
+
+        # Before it was result.size === 0, however for each version that is not correctly constructed,
+        # error messages will be concat to the array. By the time the loop reaches a version that is
+        # correctly constructe the size will stay the same as before (because it has the old error messages)
+        # so we must make sure before and after result size are the same - aka no error message was added
+        # and the version directory was correctly constructed
```
ndushay: again - I think both of the above comments add little to no value.

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+    MISSING_DIR = 1
+    EXTRA_CHILD_DETECTED = 2
+    EMPTY = 3
+    NO_SIGNATURE_CATALOG = 4
+    NO_MANIFEST_INVENTORY = 5
+    NO_XML_FILES = 6
+    VERSIONS_NOT_IN_ORDER = 7
+    FILES_IN_VERSION_DIR = 8
+
+    RESPONSE_CODE_TO_MESSAGES = {
```
ndushay: perhaps ERROR_CODE_TO_MESSAGES if they are all errors?

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+
+    attr_reader :storage_obj_path
+
+    def initialize(storage_object)
+      @storage_obj_path = storage_object.object_pathname
+      @directory_entries_hash = {}
+    end
+
+    def validate_object
+      results = []
```
ndushay: perhaps 'errors' is more appropriate than 'results' if any message added indicates error.  then the code will explain itself better and many of the comments become unnecessary?

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+        # don't bother checking the data sub dir if we didn't find the expected two subdirs
+
+        # Before it was result.size === 0, however for each version that is not correctly constructed,
+        # error messages will be concat to the array. By the time the loop reaches a version that is
+        # correctly constructe the size will stay the same as before (because it has the old error messages)
+        # so we must make sure before and after result size are the same - aka no error message was added
+        # and the version directory was correctly constructed
+
+        if before_result_size == after_result_size
+          data_dir_path = "#{version_path}/#{version_sub_dirs[0]}"
```
ndushay: I would rather you had a DATA_DIR constant and used it here:
```
DATA_DIR = 'data'.freeze # up high
data_dir_path = "#{version_path}/#{DATA_DIR}"
```

OR

```
data_dir_path = #{version_path}/#{EXPECTED_VERSION_SUB_DIRS[0]}
```
because in the suggestions above, it is clearer you getting the name from the expected constant.

using a position seems a bit more indirect that we'd desire.

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+        end
+      end
+
+      results
+    end
+
+    def check_sub_dirs(sub_dirs, version, required_sub_dirs)
+      results = []
+      sub_dir_count = sub_dirs.size
+      if sub_dir_count == 2
```
ndushay: i think here you want to substitute:  `required_sub_dirs.size` for hardcoded constant `2` -- because in this context, it could be any number, no??

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+      directory_entries(path)[:dirs]
+    end
+
+    def files_in_dir(path)
+      directory_entries(path)[:files]
+    end
+
+    def found_unexpected(array, version, required_sub_dirs)
+      results = []
+      unexpected = (array - required_sub_dirs) # removed .pop for now
```
ndushay: the comment `removed .pop for now` makes no sense to me ... perhaps you either want a more thorough comment, or an issue or something to capture your idea?

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+      unexpected = "#{unexpected} Version: #{version}"
+      results << result_hash(EXTRA_CHILD_DETECTED, unexpected)
+      results
+    end
+
+    def missing_data(array, version, required_sub_dirs)
```
ndushay: if this is about MISSING_DIR, should the method name be missing_dir ??  (error code and method name aren't matching.

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+      results = []
+      results << result_hash(INCORRECT_DIR) unless array == required_sub_dirs
+      results
+    end
+
+    def result_hash(response_code, addl=nil)
+      { response_code => result_code_msg(response_code, addl) }
+    end
+
+    def result_code_msg(response_code, addl=nil)
```
ndushay: error_code_msg if you go with the name change ...

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,196 @@
...
+      format(RESPONSE_CODE_TO_MESSAGES[response_code], addl: addl)
+    end
+
+    def check_manifest_files(dir, version)
```
ndushay: would this be more correctly named `check_required_manifest_files` ?

.rubocop.yml
```diff
@@ -27,6 +27,11 @@ Metrics/LineLength:
     - 'lib/moab/file_manifestation.rb'
     - 'spec/unit_tests/moab/bagger_spec.rb' # remove after PR #52 merged
     - 'spec/unit_tests/moab/file_group_difference_spec.rb' # remove after PR #50 merged
+    - 'spec/unit_tests/moab/storage_object_validator_spec.rb'
+
+# Max was 8
```
ndushay: think of these comments more as "why" than "what".  Like "code was readable with higher complexity allowed"

.rubocop.yml
```diff
@@ -41,6 +46,14 @@ Naming/FileName:
 
 # --- Style ---
 
+# Pointless change
+Style/FormatStringToken:
+  Enabled: false
+  
+# Had to turn this off for many files
```
ndushay: again - "why?".    "Loops were easier to understand without favoring next statements" ?

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,198 @@
+require 'moab'
+
+# Shameless gree: repetitious code included.
```
ndushay: I know I told you to put this here, but given that you have a class comment ... perhaps it's part of the class comment, not the module comment?

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,198 @@
...
+      errors
+    end
+
+    # TODO: Figure out which methods should be public
+
+    private
+
+    def check_correctly_named_version_dir
+      errors = []
+      sub_dirs(storage_obj_path).each do |version_dir|
```
ndushay: is this not version_directories?

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,198 @@
...
+      errors.concat check_sequential_version_dirs if errors.empty?
+      errors.concat check_correctly_formed_moab if errors.empty?
+      errors
+    end
+
+    # TODO: Figure out which methods should be public
+
+    private
+
+    def check_correctly_named_version_dir
```
ndushay: should method name be plural as you're checking all the version dirs?

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,198 @@
...
+      errors = []
+
+      version_directories.each_with_index do |dir_name, index|
+        expected_vers_num = index + 1 # version numbering starts at 1, array indexing starts at 0
+        begin
+          if dir_name[1..-1].to_i != expected_vers_num
+            errors << result_hash(VERSIONS_NOT_IN_ORDER, version_directories)
+            break
+          end
+        rescue ArgumentError
```
ndushay: I'm still not clear on how you would get an argument error, nor do you seem to be, as you don't have a test for this rescue clause, per coveralls.

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,198 @@
...
+    def check_sub_dirs(sub_dirs, version, required_sub_dirs)
+      errors = []
+      sub_dir_count = sub_dirs.size
+      if sub_dir_count == required_sub_dirs.size
+        errors.concat expected_dirs(sub_dirs, version, required_sub_dirs)
+      elsif sub_dir_count > required_sub_dirs.size
+        errors.concat found_unexpected(sub_dirs, version, required_sub_dirs)
+      elsif sub_dir_count < required_sub_dirs.size
+        errors.concat missing_dir(sub_dirs, version, required_sub_dirs)
+      elsif sub_dir_count.zero?
```
ndushay: coveralls says you have no test that uses this code

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,198 @@
...
+
+    # This method removes the implicit '.' and '..' directories.
+    # Returns an array of strings.
+    def directory_entries(path)
+      @directory_entries_hash[path] ||=
+        begin
+          ret_val = { files: [], dirs: [] }
+          (Dir.entries(path).sort - IMPLICIT_DIRS).each do |child|
+            if FileTest.file? child
+              ret_val[:files] << child
```
ndushay: coveralls says you have no test that uses this code.

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+        # v0006
+        expect(verification_array[6]).to eq(1=>"Missing directory: [\"content\", \"metadata\"] Version: v0006")
+      end
+      it 'has no items' do
+        # v0006
+        expect(verification_array[7]).to eq(6=>"Version: v0006 Missing all required metadata files")
+      end
+    end
+  end
+  describe '#validation_errors' do
```
ndushay: one might argue for one context block for '#validation_errors' and 3 diff tests.  The unique values for vars (which are only used by one test apiece) could be declared inside the it blocks;  let could be used for the vars that are the same across the 3 specs.

This would make the spec more compact and probably easier to understand.

lib/moab/storage_object_validator.rb
SaravShah: Ah yeah tomorrow you'll have to show me how to use coveralls to see what parts of my code is not tested and is tested. You get an ArgumentError when something cannot be converted to an integer from ``` if dir_name[1..-1].to_i != expected_vers_num ```

SaravShah: So i'll change the error code to `VERSION_DIR_BAD_FORMAT` (used to make sure the version directories are correctly named)

SaravShah: Switched it from  ``` to_i ``` to ``` Integer() ```

.rubocop.yml
```diff
@@ -27,6 +27,11 @@ Metrics/LineLength:
     - 'lib/moab/file_manifestation.rb'
     - 'spec/unit_tests/moab/bagger_spec.rb' # remove after PR #52 merged
     - 'spec/unit_tests/moab/file_group_difference_spec.rb' # remove after PR #50 merged
+    - 'spec/unit_tests/moab/storage_object_validator_spec.rb'
+
+# Code was readable with allowing higher perceived complexity
```
ndushay: 👍

.rubocop.yml
```diff
@@ -41,6 +46,14 @@ Naming/FileName:
 
 # --- Style ---
 
+# Pointless change
+Style/FormatStringToken:
+  Enabled: false
+  
+# Loops were easier to understand without using 'next' statement
```
ndushay: 👍

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,182 @@
+require 'moab'
+
+module Moab
+  # Given a druid path, are the contents actually a well-formed Moab?
+  # Shameless green: repetitious code included.
```
ndushay: 👍

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+      it 'has missing manifest files' do
+        # v0005
+        expect(verification_array[5]).to eq(6=>"Version: v0005 Missing all required metadata files")
+      end
+      it 'has missing content directory' do
+        # v0006
+        expect(verification_array[6]).to eq(1=>"Missing directory: [\"content\", \"metadata\"] Version: v0006")
+      end
+      it 'has no requied items' do
```
ndushay: OMG typo. "requied".

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,182 @@
...
+          nil
+        else
+          errors << result_hash(VERSION_DIR_BAD_FORMAT)
+        end
+      end
+      errors
+    end
+
+    # This method will be called only if the version directories are correctly
+    # named
```
ndushay: OMG unnecessary line break.

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      expect(storage_obj_validator1.validation_errors).to be_kind_of Array
+    end
+
+    it 'has item in the array that is the expected structure' do
+      expect(verification_array[0]).to be_kind_of Hash
+    end
+    context 'under version directory' do
+      it 'has missing items' do
+        # v0001
+        expect(verification_array[0]).to eq(1=>"Missing directory: [\"data\", \"manifests\"] Version: v0001")
```
jmartin-sul: feels to me like it'd be clearer to reference the error code in the result hash by its constant name instead of the underlying int val, e.g. 

```ruby
expect(verification_array[0]).to eq(StorageObjectValidator::MISSING_DIR => "Missing directory: [\"data\", \"manifests\"] Version: v0001")
```

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,181 @@
...
+    def check_correctly_formed_moabs
+      errors = []
+      version_directories.each do |version_dir|
+        version_path = "#{storage_obj_path}/#{version_dir}"
+        version_sub_dirs = sub_dirs(version_path)
+        before_result_size = errors.size
+        errors.concat check_sub_dirs(version_sub_dirs, version_dir, EXPECTED_VERSION_SUB_DIRS)
+        after_result_size = errors.size
+
+        if before_result_size == after_result_size
```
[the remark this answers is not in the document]
SaravShah: Yeah i will add that comment. Earlier i had a huge comment that was trying to explain it, but then i took it out. I'll add that comment to make it easier. It's kind of a funny way to do this and am wondering if there is a better way

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,181 @@
...
+          data_dir_path = "#{version_path}/#{EXPECTED_VERSION_SUB_DIRS[0]}"
```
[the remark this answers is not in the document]
SaravShah: Yeah, probably better for readability.

lib/moab/storage_object_validator.rb
```diff
@@ -0,0 +1,181 @@
...
+          (Dir.entries(path).sort - IMPLICIT_DIRS).each do |child|
+            dirs << child
+          end
+          dirs
+        end
+    end
+
+    def sub_dirs(path)
+      directory_entries(path)
+    end
```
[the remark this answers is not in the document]
SaravShah: I kinda left this like this just because I didn't incorporate the ``` [ ] content and metadata directories each contain only files ```.  
```
def directory_entries(path)
      @directory_entries_hash[path] ||=
        begin
          ret_val = { files: [], dirs: [] }
          (Dir.entries(path).sort - IMPLICIT_DIRS).each do |child|
            if FileTest.file? child
              ret_val[:files] << child
            else
              ret_val[:dirs] << child
            end
          end
          ret_val
        end
    end

    def sub_dirs(path)
      directory_entries(path)[:dirs]
    end

    def files_in_dir(path)
      directory_entries(path)[:files]
    end
```


 (you refactored that part, but i removed it for this PR. I created another ticket for that, so i could get this merged and a new gem release that was demoable. Also, the ``` File.file? ``` part of that code wasn't working properly so I just decided to work on that in another ticket ... )

lib/moab/storage_object_validator.rb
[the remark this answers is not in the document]
jmartin-sul: i think it's a reasonable approach (or at least, i'm also unable to think of a cleaner approach)

Issue #74, opened 2017-10-28
(MV) Version directories are correct format 'v00xx'


SaravShah:
Maybe use regex to test if version directories are in correct format. "v0001" "v0010" "v00xx"?

SaravShah:
in favor of #72

PR #75, opened 2017-10-30
74 ver dir format

SaravShah:
Branched off  [WIP] First pass Storage object validator #72.
- Added new spec fixture for test
- Logic for checking version directory name w/ regex
- Will need to rebase if / when #72 is merged
- Changes are the last two commits in commit history ( 9f632bd, 6bcb250 )

connect #74


SaravShah:
In favor of #72

PR #80, opened 2017-11-03
Check whether druid matches most recent version object_id

### LINKED ISSUE #71: (D2D) check druid from most recent manifestInv.xml against directory name
identify the most recent version, and positively identify the druid (read most recent manifestInventory.xml for objectId; match against druidtree path - i.e. does the path /aa/123/bb/1234/aa123bb1234 really contain a Moab object of the druid aa123bb1234? ).

tallenaz:
Fixes #71


lib/moab/storage_object_validator.rb
```diff
@@ -50,6 +52,16 @@ def validation_errors
       errors
     end
 
+    def identify_druid
+      errors = []
+      if druid_from_filepath == object_id_from_manifest_inventory
+        nil
+      else
+        errors << result_hash(DRUID_MISMATCH)
+      end
```
jmartin-sul: how about:
```ruby
errors << result_hash(DRUID_MISMATCH) unless druid_from_filepath == object_id_from_manifest_inventory
```

spec/fixtures/storage_root01/moab_storage_trunk/bj/102/hs/9687/bj102hs9687/v0003/manifests/manifestInventory.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- objectId matches druid implied in file path -->
```
jmartin-sul: the comment is in this file (for druid `bj102hs9687`), but it appears that the druid being used in the specs is `bz514sm9647`.  comment here is still totally valid, of course, but i guess i'd have expected the new test and the comment to correspond?

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -84,4 +84,23 @@
...
+    let(:different_druid_path) { 'spec/fixtures/storage_root01/moab_storage_trunk/jj/925/bx/9565/jj925bx9565' }
+    let(:storage_obj2) { Moab::StorageObject.new(different_druid, different_druid_path) }
+    let(:storage_obj_validator2) { described_class.new(storage_obj2) }
+
+    it 'returns empty errors array when druid is the same' do
+      expect(storage_obj_validator1.identify_druid).to eq []
+    end
+
+    it 'returns the right error code when druid is not the same' do
+      expect(storage_obj_validator2.identify_druid).to eq [{ 9=>'manifestInventory object_id does not match druid' }]
```
jmartin-sul: feels to me like it'd be slightly clearer to use `Moab::StorageObjectValidator::DRUID_MISMATCH` instead of the underlying int code.

lib/moab/storage_object_validator.rb
```diff
@@ -176,5 +188,21 @@ def check_required_manifest_files(dir, version)
       errors << result if result
       errors
     end
+
+    def druid_from_filepath
+      "druid:#{File.basename(storage_obj_path)}"
+    end
+
+    def latest_druid_version
+      sub_dirs(storage_obj_path).last
```
SaravShah: why did you have to add `.last`? could you used `version_directories.last` here?

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -84,4 +84,23 @@
...
+  context '#identify_druid' do
+    let(:same_druid) { 'bz514sm9647' }
+    let(:same_druid_path) { 'spec/fixtures/storage_root01/moab_storage_trunk/bz/514/sm/9647/bz514sm9647' }
+    let(:storage_obj1) { Moab::StorageObject.new(same_druid, same_druid_path) }
+    let(:storage_obj_validator1) { described_class.new(storage_obj1) }
+    let(:different_druid) { 'jj925bx9565' }
+    let(:different_druid_path) { 'spec/fixtures/storage_root01/moab_storage_trunk/jj/925/bx/9565/jj925bx9565' }
+    let(:storage_obj2) { Moab::StorageObject.new(different_druid, different_druid_path) }
+    let(:storage_obj_validator2) { described_class.new(storage_obj2) }
+
```
SaravShah: i think for this it might be easier to read if you declare the variables inside the `it ..... do.......end ` block rather than declaring all the variables at the top... but if you don't agree then i'm ok with how it is

lib/moab/storage_object_validator.rb
SaravShah: Add if you could use `version_directories.last` then would you need this method ?

tallenaz:
I was able to 
- get `identify_druid` down to one line
- correct comments in manifestInv stubs
- refer to the error in results hash semantically instead of via numeric code
- create and test a Stanford specific validator that inherits from the moab validator
- get rid of a method via interpolating `version_directories.last` straight into `latest_manifest_inventory`
I'm wincing at the duplication of `result_hash` and `error_code_msg`

tallenaz:
I’d like to write a spec for the override.

lib/stanford/storage_object_validator.rb
```diff
@@ -0,0 +1,63 @@
+require 'moab'
+
+module Stanford
+  # druids are Stanford specific entities
+  class StorageObjectValidator < Moab::StorageObjectValidator
+    # error codes
+    EXPECTED_DATA_SUB_DIRS = ["content", "metadata"].freeze
```
ndushay: You shouldn't need to re-declare class constants -- you should be able to use them without redeclaring.

lib/stanford/storage_object_validator.rb
```diff
@@ -0,0 +1,63 @@
...
+      MISSING_DIR => "Missing directory: %{addl}",
+      EXTRA_CHILD_DETECTED => "Unexpected item in path: %{addl}",
+      VERSION_DIR_BAD_FORMAT => "Version directory name not in 'v00xx' format",
+      FILES_IN_VERSION_DIR => "Top level should contain only sequential version directories. Also contains files: %{addl}",
+      NO_SIGNATURE_CATALOG => "Version: %{addl} Missing signatureCatalog.xml",
+      NO_MANIFEST_INVENTORY => "Version: %{addl} Missing manifestInventory.xml",
+      NO_XML_FILES => "Version: %{addl} Missing all required metadata files",
+      VERSIONS_NOT_IN_ORDER => "Should contain only sequential version directories. Current directories: %{addl}",
+      DRUID_MISMATCH => "manifestInventory object_id does not match druid"
+    }.freeze
```
ndushay: I'm wondering if all the duplicated constant stuff above is simply because we should take `.freeze` off of ERROR_CODE_TO_MESSAGES in the super class.  Then this could be

```
ERROR_CODE_TO_MESSAGES[DRUID_MISMATCH] = "manifestInventory object_id does not match druid"
```

lib/stanford/storage_object_validator.rb
```diff
@@ -0,0 +1,63 @@
...
+    INCORRECT_DIR = 0
+    MISSING_DIR = 1
+    EXTRA_CHILD_DETECTED = 2
+    VERSION_DIR_BAD_FORMAT = 3
+    NO_SIGNATURE_CATALOG = 4
+    NO_MANIFEST_INVENTORY = 5
+    NO_XML_FILES = 6
+    VERSIONS_NOT_IN_ORDER = 7
+    FILES_IN_VERSION_DIR = 8
+    DRUID_MISMATCH = 9
```
ndushay: if you wanted to get fancy, perhaps you could do

```
# note that next line assumes keys are in ascending numerical order
DRUID_MISMATCH = ERROR_CODE_TO_MESSAGES.keys.last + 1
```

not sure if that's better than simply hard-coding a 9.

lib/stanford/storage_object_validator.rb
```diff
@@ -0,0 +1,63 @@
...
+      NO_SIGNATURE_CATALOG => "Version: %{addl} Missing signatureCatalog.xml",
+      NO_MANIFEST_INVENTORY => "Version: %{addl} Missing manifestInventory.xml",
+      NO_XML_FILES => "Version: %{addl} Missing all required metadata files",
+      VERSIONS_NOT_IN_ORDER => "Should contain only sequential version directories. Current directories: %{addl}",
+      DRUID_MISMATCH => "manifestInventory object_id does not match druid"
+    }.freeze
+
+    def validation_errors
+      errors = []
+      errors.concat identify_druid
```
ndushay: I *think* this might need to be << if it's a single entry ???  @jmartin-sul knows better.

lib/stanford/storage_object_validator.rb
```diff
@@ -0,0 +1,63 @@
...
+      super if errors.empty?
```
ndushay: I think John was suggesting something like

```
errors = []
# to get ALL errors -- remove this comment in actual code
errors.concat super
errors << identify_druid # if i'm right about .concat vs. <<
```

Note that you don't need separate line "errors" as   errors << obj   returns the array.

lib/stanford/storage_object_validator.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    def identify_druid
+      druid_from_filepath == object_id_from_manifest_inventory ? [] : [result_hash(DRUID_MISMATCH)]
+    end
+
+    private
+
+    def result_hash(response_code, addl=nil)
+      { response_code => error_code_msg(response_code, addl) }
+    end
```
ndushay: are `result_hash` and `error_code_msg` private methods the same as in superclass?  If so, shouldn't need to redeclare them.

spec/unit_tests/stanford/storage_object_validator_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        expect(storage_obj_validator1.identify_druid).to eq []
+      end
+
+      it 'returns the right error code when druid is not the same' do
+        expect(storage_obj_validator2.identify_druid).to eq [
+          { Stanford::StorageObjectValidator::DRUID_MISMATCH=>'manifestInventory object_id does not match druid' }
+        ]
+      end
+    end
+  end
```
ndushay: add test:  it 'returns errors from calling super Moab::validation_errors'

tallenaz:
- Stanford::StorageObjectValidator#validation_errors now adds to errors instead of short-circuiting the super call on a D2D error.
- The Stanford class is much less duplicative of the Moab::StorageObjectValidator
- There's a test to make sure the Stanford validator returns non-Stanford specific errors by running it on a problematic druid without a D2D error, and expecting there to be errors.

lib/moab/storage_object_validator.rb
```diff
@@ -50,6 +38,21 @@ def validation_errors
       errors
     end
 
+    def self.error_code_to_messages
+      @error_code_to_messages ||=
+        {
```
ndushay: maybe a comment indicating Stanford::MOV is expecting the keys to be in ascending numerical order?

lib/stanford/storage_object_validator.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    # TODO: test to make sure constants don't collide on underlying int vals?
+    # keep from stepping on previously defined error code constants.
+    DRUID_MISMATCH = superclass.error_code_to_messages.keys.max + 1
+
+    def validation_errors
+      errors = []
+      errors.concat super
+      errors << identify_druid
+      errors.flatten
```
ndushay: i think the idea of using `.concat` for multiple values is to avoid having to use flatten.  But no biggie.

tallenaz:
- added a comment clarifying how Stanford validator expects keys
- Stanford validation_errors uses a single call to concat to build errors

lib/moab/storage_object_validator.rb
jmartin-sul: i think that might not be needed, since the subclass does `superclass.error_code_to_messages.keys.max` to get the offset for its error codes.

lib/stanford/storage_object_validator.rb
jmartin-sul: i would also vote for `errors.concat identify_druid`.  if only because we're otherwise consistent about that elsewhere (but it also just reads a little nicer as one step, imho).

SaravShah:
looks good with the changes!

lib/stanford/storage_object_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def identify_druid
+      druid_from_filepath == object_id_from_manifest_inventory ? [] : [result_hash(DRUID_MISMATCH)]
+    end
+
+    # the Stanford validator expects keys to be in ascending numerical order
+    def self.error_code_to_messages
```
ndushay: this should be private here and in moab::SOV

tallenaz:
@ndushay when we make `error_code_to_messages` private, the test suite throws multiple instances of `NoMethodError`:

`Failure/Error: DRUID_MISMATCH = superclass.error_code_to_messages.keys.max + 1`

ndushay:
ah - ok - so let's leave error_code_to_messages public.  I'll re-approve this PR, leaving only @jmartin-sul to do it.

PR #86, opened 2017-11-07
Document Stanford::StorageOjbectValidator in README

### LINKED ISSUE #84: README: Explain how to use D2D code
- Update usage information in the stanford-specific area of README

tallenaz:
Fixes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,11 +57,11 @@ size = moab.size
 #### Validate if Moab Object is Well-Formed
 
 ```ruby
-moab = Moab::StorageObject.new(druid, object_dir)
+moab = Moab::StorageObject.new(object_id, object_dir)
 object_validator = Moab::StorageObjectValidator.new(moab)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,11 +57,11 @@ size = moab.size
 #### Validate if Moab Object is Well-Formed
 
 ```ruby
-moab = Moab::StorageObject.new(druid, object_dir)
+moab = Moab::StorageObject.new(object, object_dir)
```
jmartin-sul: maybe `object_id` instead of `object`?  this makes me think it'll be a more complex instantiated object, but it's a string id, right?

good catch on not making it `druid` in the non-stanford case, though... maybe we should have another PR that makes that change across this whole README?

tallenaz: Agreed that `object_id` is clearer. Made that change, and ticketed https://github.com/sul-dlss/moab-versioning/issues/87

PR #90, opened 2017-11-09
Run identify_druid only if we know there is a moab with the correct dirs

SaravShah:
- Only run identify_druid if we know there is a manifest directory / correctly constructed Moab
- added correct test to spec/unit_tests/stanford

### DIFF
--- a/lib/stanford/storage_object_validator.rb
+++ b/lib/stanford/storage_object_validator.rb
@@ -11,7 +11,8 @@ module Stanford
     def validation_errors
       errors = []
       errors.concat super
-      errors.concat identify_druid
+      errors.concat(identify_druid) if errors.empty?
+      errors
     end
 
     def identify_druid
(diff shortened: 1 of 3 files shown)


lib/stanford/storage_object_validator.rb
```diff
@@ -11,7 +11,9 @@ class StorageObjectValidator < Moab::StorageObjectValidator
     def validation_errors
       errors = []
       errors.concat super
-      errors.concat identify_druid
+      errors.concat(identify_druid) if errors.empty?
+      errors
+      # errors.concat identify_druid if errors.empty?
```
ndushay: get rid of this comment?

SaravShah:
ticket created #91

lib/stanford/storage_object_validator.rb
```diff
@@ -11,7 +11,8 @@ class StorageObjectValidator < Moab::StorageObjectValidator
     def validation_errors
       errors = []
       errors.concat super
-      errors.concat identify_druid
+      errors.concat(identify_druid) if errors.empty?
+      errors
```
ndushay: I do not care if this is addressed now (or ever) but ... do we need L15?

jmartin-sul: IIRC, line 14 will return `nil` if the if condition isn't met...

yeah:
```ruby
[1] pry(main)> 'hello' if true
=> "hello"
[2] pry(main)> 'hello' if false
=> nil
[3] pry(main)> 
```

Issue #79, opened 2017-11-02
(MV) content and metadata directories each contain only files.


SaravShah:
- create RSpec tests for this
- Each directory should only contain files. something like this:
``` 
def directory_entries(path)
      @directory_entries_hash[path] ||=
        begin
          ret_val = { files: [], dirs: [] }
          (Dir.entries(path).sort - IMPLICIT_DIRS).each do |child|
            if FileTest.file? child
              ret_val[:files] << child
            else
              ret_val[:dirs] << child
            end
          end
          ret_val
        end
    end

    def sub_dirs(path)
      directory_entries(path)[:dirs]
    end

    def files_in_dir(path)
      directory_entries(path)[:files]
    end

```

ndushay:
addressed by #92

PR #92, opened 2017-11-09
Metadata folder only contains files

SaravShah:
Only checks metadata folder for files - though now that i'm thinking about it, maybe this should wait till after Tommy's PR (about checking existence of content directory) is merged because once his logic is implemented, if there is a content directory then i can change my code to implement that.  (Check if the content directory only contains files)
connects #79 
connects #94


lib/moab/storage_object_validator.rb
```diff
@@ -22,6 +23,7 @@ class StorageObjectValidator
     NO_XML_FILES = 6
     VERSIONS_NOT_IN_ORDER = 7
     FILES_IN_VERSION_DIR = 8
+    DIRS_DETECTED = 9
```
ndushay: I *think* this may impact Stanford::StorageObjectValidator, in that it declares one more error code constant ...

lib/moab/storage_object_validator.rb
```diff
@@ -98,6 +102,13 @@ def check_correctly_formed_moabs
           errors.concat check_sub_dirs(data_sub_dirs, version_dir, EXPECTED_DATA_SUB_DIRS)
           errors.concat check_required_manifest_files(version_path, version_dir)
         end
+        if before_result_size == errors.size
```
ndushay: can these statements be added to the if block above?  if not, then isn't the if condition wrong?

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -97,6 +101,7 @@
       storage_obj_validator = described_class.new(storage_obj)
       verification_array = storage_obj_validator.validation_errors
       expect(verification_array).to include(Moab::StorageObjectValidator::INCORRECT_DIR => "Incorrect items in path")
+
```
ndushay: where did this blank line come from?

lib/moab/storage_object_validator.rb
SaravShah: changed the order - i think it should be fine now

lib/moab/storage_object_validator.rb
SaravShah: Well i guess here, In the first if block 
``` 
if before_result_size == after_result_size
          data_dir_path = "#{version_path}/#{DATA_DIR_NAME}"
          data_sub_dirs = sub_dirs(data_dir_path)
          errors.concat check_sub_dirs(data_sub_dirs, version_dir, EXPECTED_DATA_SUB_DIRS)
          errors.concat check_required_manifest_files(version_path, version_dir)
        end
```

I want this to run if the sizes are equal, meaning nothing has been added to the errors array. Now if this if loop is executed then the errors array will either stay the same size or get bigger. Now if it stays the same size i then want the next if loop to run ( ``` if before_result_size == errors.size
          x = []
          metadata_dir_path = "#{version_path}/#{DATA_DIR_NAME}/#{METADATA_DIR_NAME}"
          metadata_sub_dir = sub_dirs(metadata_dir_path)
          metadata_sub_dir.each { |item| x << File.directory?("#{metadata_dir_path}/#{item}") }
          errors << result_hash(DIRS_DETECTED) if x.include?(true)
        end ``` ) but if the errors array is bigger then i don't want the 2nd if loop to run

SaravShah: I'm sure there is a better way of coding this then 2 if loops.

SaravShah: i'm trying to think of a better way to do this

lib/moab/storage_object_validator.rb
jmartin-sul: might be misunderstanding, but should be ok because of the auto-offset in `Stanford::StorageObjectValidator` and because we don't ever serialize those values and read them back later (and so they just have to stay consistent for a single validation execution).

https://github.com/sul-dlss/moab-versioning/blob/4016e48235d2cd152161701e02b8319937ba28e1/lib/stanford/storage_object_validator.rb#L9

SaravShah: oh right... that makes sense

ndushay: Thanks @jmartin-sul - I should have looked it up myself.  I'm thinking we should have a test for this (and there's a TODO comment to that effect) so I created #310 with that in mind.

lib/moab/storage_object_validator.rb
ndushay: Just noting that @SaravShah and I spoke F2F and are thinking an approach like this might be clearer:

```
if before_result_size == after_result_size
      data_dir_path = "#{version_path}/#{DATA_DIR_NAME}"
      data_sub_dirs = sub_dirs(data_dir_path)
      errors.concat check_sub_dirs(data_sub_dirs, version_dir, EXPECTED_DATA_SUB_DIRS)
      errors.concat check_required_manifest_files(version_path, version_dir)
      if before_result_size == errors.size
          x = []
          metadata_dir_path = "#{version_path}/#{DATA_DIR_NAME}/#{METADATA_DIR_NAME}"
          metadata_sub_dir = sub_dirs(metadata_dir_path)
          metadata_sub_dir.each { |item| x << File.directory?("#{metadata_dir_path}/#{item}") }
          errors << result_hash(DIRS_DETECTED) if x.include?(true)
        end  
end
```  

It might make sense to make the inner conditional code into a well-named method ...

Note that the current implementation is NOT recomputing after_result_size so won't it always do both conditionals, which is incorrect (there should be a test for this in both directions, no?).

SaravShah:
Added 2 RSpec tests and reorganized the tests to make it more readable? Let me know if you guys like what I did. 
Also took Naomi's suggestion and made a nested if condition and created extra methods.

.rubocop_todo.yml
```diff
@@ -440,6 +440,8 @@ RSpec/MessageSpies:
 # Offense count: 131
 RSpec/MultipleExpectations:
   Max: 14
+RSpec/NestedGroups:
+  Max: 4 # previously max 3
```
ndushay: comment explaining "why" ... doesn't explain "why".   Maybe "tests readable with this" or possibly better:  exclude the particular failing file, and comment "why" for that specific file.

lib/moab/storage_object_validator.rb
```diff
@@ -93,13 +96,31 @@ def check_correctly_formed_moabs
...
+          errors.concat check_expected_data_sub_dir(version_path, version_dir)
+          if after_result_size == errors.size
+            errors.concat check_metadata_dir_file_only(version_path)
+          end
         end
       end
+      errors
+    end
+
+    def check_expected_data_sub_dir(version_path, version_dir)
```
ndushay: super picky ... but isn't this checking expected_data_sub_dirS  ?

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,136 @@
     let(:verification_array) { storage_obj_validator1.validation_errors }
 
     it 'returns correct data structure' do
-      expect(storage_obj_validator1.validation_errors).to be_kind_of Array
+      expect(verification_array).to be_kind_of Array
```
ndushay: 👍

lib/moab/storage_object_validator.rb
```diff
@@ -93,13 +96,31 @@ def check_correctly_formed_moabs
...
         # run the following checks if this version dir passes check_sub_dirs, even if some prior version dirs didn't
         if before_result_size == after_result_size
-          data_dir_path = "#{version_path}/#{DATA_DIR_NAME}"
-          data_sub_dirs = sub_dirs(data_dir_path)
-          errors.concat check_sub_dirs(data_sub_dirs, version_dir, EXPECTED_DATA_SUB_DIRS)
-          errors.concat check_required_manifest_files(version_path, version_dir)
+          errors.concat check_expected_data_sub_dirs(version_path, version_dir)
+          if after_result_size == errors.size
+            errors.concat check_metadata_dir_file_only(version_path)
+          end
```
ndushay: I think this erroneously removes "check_required_manifest_files" check, no?   If you put it back, you can do #94 as well ...

SaravShah: 💯 👍

lib/moab/storage_object_validator.rb
```diff
@@ -92,14 +95,32 @@ def check_correctly_formed_moabs
...
         # run the following checks if this version dir passes check_sub_dirs, even if some prior version dirs didn't
+        errors.concat check_required_manifest_files(version_path, version_dir) if before_result_size == after_result_size
         if before_result_size == after_result_size
-          data_dir_path = "#{version_path}/#{DATA_DIR_NAME}"
-          data_sub_dirs = sub_dirs(data_dir_path)
-          errors.concat check_sub_dirs(data_sub_dirs, version_dir, EXPECTED_DATA_SUB_DIRS)
-          errors.concat check_required_manifest_files(version_path, version_dir)
+          errors.concat check_expected_data_sub_dirs(version_path, version_dir)
+          if after_result_size == errors.size
+            errors.concat check_metadata_dir_file_only(version_path)
```
ndushay: well, shoot.  `check_metadata_dir_fileSSSSSSS_only`. (do you speak parseltongue?)

lib/moab/storage_object_validator.rb
```diff
@@ -92,14 +95,32 @@ def check_correctly_formed_moabs
         errors.concat check_sub_dirs(version_sub_dirs, version_dir, EXPECTED_VERSION_SUB_DIRS)
         after_result_size = errors.size
         # run the following checks if this version dir passes check_sub_dirs, even if some prior version dirs didn't
+        errors.concat check_required_manifest_files(version_path, version_dir) if before_result_size == after_result_size
         if before_result_size == after_result_size
```
ndushay: now this has to be errors.size, not after_result_size, as the previous line can change errors.size  Probably you should just always do errors.size if you can.  :-P

lib/moab/storage_object_validator.rb
```diff
@@ -49,6 +51,7 @@ def self.error_code_to_messages
           NO_SIGNATURE_CATALOG => "Version: %{addl} Missing signatureCatalog.xml",
           NO_MANIFEST_INVENTORY => "Version: %{addl} Missing manifestInventory.xml",
           NO_XML_FILES => "Version: %{addl} Missing all required metadata files",
+          DIRS_DETECTED => "Should only contain files, but directories were present",
```
ndushay: it would be good to indicate which directory we are talking about here

lib/moab/storage_object_validator.rb
```diff
@@ -49,6 +51,7 @@ def self.error_code_to_messages
           NO_SIGNATURE_CATALOG => "Version: %{addl} Missing signatureCatalog.xml",
           NO_MANIFEST_INVENTORY => "Version: %{addl} Missing manifestInventory.xml",
           NO_XML_FILES => "Version: %{addl} Missing all required metadata files",
```
ndushay: is this about data/METADATA or about MANIFESTS ?  Or both?  Does *metadata* have required files?  Are they all XML?

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,137 @@
...
+          # v0005
+          expect(verification_array[4]).to eq(Moab::StorageObjectValidator::NO_XML_FILES =>"Version: v0005 Missing all required metadata files")
+          # v0007
+          expect(verification_array[8]).to eq(Moab::StorageObjectValidator::NO_XML_FILES =>"Version: v0007 Missing all required metadata files")
+        end
+        it 'has missing content directory' do
+          # v0006
+          expect(verification_array[7]).to eq(Moab::StorageObjectValidator::MISSING_DIR =>"Missing directory: [\"content\", \"metadata\"] Version: v0006")
+        end
+        it 'has no required items' do
```
ndushay: here's the test:  does METADATA have required files?

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,137 @@
...
+        storage_obj = Moab::StorageObject.new(druid, druid_path)
+        storage_obj_validator = described_class.new(storage_obj)
+        verification_array = storage_obj_validator.validation_errors
+        it 'does not have signatureCatalog.xml' do
+          expect(verification_array).to include(Moab::StorageObjectValidator::NO_SIGNATURE_CATALOG => "Version: v0001 Missing signatureCatalog.xml")
+        end
+
+        it 'does not have manifestInventory.xml' do
+          expect(verification_array).to include(Moab::StorageObjectValidator::NO_MANIFEST_INVENTORY => "Version: v0002 Missing manifestInventory.xml")
+        end
```
ndushay: Is this where we should be testing for "all" required xml files?

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,137 @@
...
-    end
-    context 'under manifest directory' do
-      druid = 'cc000cc0000'
-      druid_path = 'spec/fixtures/bad_root01/bad_moab_storage_trunk/cc/000/cc/0000/cc000cc0000'
-      storage_obj = Moab::StorageObject.new(druid, druid_path)
-      storage_obj_validator = described_class.new(storage_obj)
-      verification_array = storage_obj_validator.validation_errors
-      it 'does not have signatureCatalog.xml' do
-        expect(verification_array).to include(Moab::StorageObjectValidator::NO_SIGNATURE_CATALOG => "Version: v0001 Missing signatureCatalog.xml")
+      it 'does not call #check_metadata_dir_file_only because moab does not have the expected data sub dirs' do
```
ndushay: don't forget to update this description to fileSSSSSS too.

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,137 @@
...
+        druid = 'ff000ff0000'
+        druid_path = 'spec/fixtures/bad_root01/bad_moab_storage_trunk/ff/000/ff/0000/ff000ff0000'
+        storage_obj = Moab::StorageObject.new(druid, druid_path)
+        storage_obj_validator = described_class.new(storage_obj)
+        expect(storage_obj_validator).not_to receive(:check_metadata_dir_file_only)
+        storage_obj_validator.validation_errors
       end
```
ndushay: do we now need tests to show what happens when manifest dir passes, vs. when manifest dir fails?

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,137 @@
...
+        expect(verification_array.empty?).to eq true
+      end
+      it 'calls #check_expected_data_sub_dirs if moab has the expected version sub dirs' do
+        expect(storage_obj_validator).to receive(:check_expected_data_sub_dirs).and_return([])
+        storage_obj_validator.validation_errors
+      end
+      it 'calls #check_metadata_dir_file_only if moab has the expected data sub dirs' do
+        expect(storage_obj_validator).to receive(:check_metadata_dir_file_only).and_return([])
+        storage_obj_validator.validation_errors
       end
```
ndushay: add "valid moab" test to show it calls check_manifest_hoo_ha, too?

do you want to add tests to show it calls the version dir checks, too?

lib/moab/storage_object_validator.rb
SaravShah: NO_XML_FILES is about the metadata directory. I guess instead of saying "Missing all required metadata files" It should say "Missing .xml files and contains other items"? Metadata _does_ _not_ have required files, the directory should only contain files.

lib/moab/storage_object_validator.rb
SaravShah: 👍

lib/moab/storage_object_validator.rb
SaravShah: 👍

lib/moab/storage_object_validator.rb
SaravShah: hahaha 👍

spec/unit_tests/moab/storage_object_validator_spec.rb
SaravShah: ah yea it shouldn't say required items it should say "only contains files"

spec/unit_tests/moab/storage_object_validator_spec.rb
SaravShah: manifest dir fails in line 74,78 (does not have _required_ files - sigCat.xml and manInv.xml)

SaravShah: i'll add a test for when manifest dir passes though

spec/unit_tests/moab/storage_object_validator_spec.rb
SaravShah: yup!

lib/moab/storage_object_validator.rb
```diff
@@ -19,9 +20,10 @@ class StorageObjectValidator
     VERSION_DIR_BAD_FORMAT = 3
     NO_SIGNATURE_CATALOG = 4
     NO_MANIFEST_INVENTORY = 5
-    NO_XML_FILES = 6
+    NO_XML_FILES_IN_MANIFEST_DIR = 6
```
ndushay: So here's a question:  is it an error if EITHER the "metadata"  OR  the "manifests" directory has no files?   If so, you could use a NO_FILES_IN_DIR error and supply the directory as "addl" arg. 

Are you actually checking for *xml* files?  Or just files?  The KISS principle would be just check for files.

lib/moab/storage_object_validator.rb
```diff
@@ -92,14 +95,32 @@ def check_correctly_formed_moabs
         errors.concat check_sub_dirs(version_sub_dirs, version_dir, EXPECTED_VERSION_SUB_DIRS)
         after_result_size = errors.size
         # run the following checks if this version dir passes check_sub_dirs, even if some prior version dirs didn't
+        errors.concat check_required_manifest_files(version_path, version_dir) if before_result_size == errors.size
         if before_result_size == after_result_size
```
ndushay: this line also:  if before_result_size == errors.size

lib/moab/storage_object_validator.rb
```diff
@@ -92,14 +95,32 @@ def check_correctly_formed_moabs
         errors.concat check_sub_dirs(version_sub_dirs, version_dir, EXPECTED_VERSION_SUB_DIRS)
         after_result_size = errors.size
```
ndushay: don't even compute "after_result_size" because you can't use it on line 99, and you're not using it on line 98 either ?

lib/moab/storage_object_validator.rb
```diff
@@ -92,14 +95,32 @@ def check_correctly_formed_moabs
...
+
+    def check_expected_data_sub_dirs(version_path, version_dir)
+      errors = []
+      data_dir_path = "#{version_path}/#{DATA_DIR_NAME}"
+      data_sub_dirs = sub_dirs(data_dir_path)
+      errors.concat check_sub_dirs(data_sub_dirs, version_dir, EXPECTED_DATA_SUB_DIRS)
+      errors
+    end
 
+    def check_metadata_dir_files_only(version_path)
```
ndushay: would you also like this method to check that there is at least ONE file?

SaravShah: I added the test for Metadata dir empty. I think if there is at least ONE file - we won't get any error codes.

SaravShah:
I'm feeling good on this one

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +98,36 @@ def check_correctly_formed_moabs
...
-          data_dir_path = "#{version_path}/#{DATA_DIR_NAME}"
-          data_sub_dirs = sub_dirs(data_dir_path)
-          errors.concat check_sub_dirs(data_sub_dirs, version_dir, EXPECTED_DATA_SUB_DIRS)
-          errors.concat check_required_manifest_files(version_path, version_dir)
+        errors.concat check_required_manifest_files(version_path, version_dir) if version_error_count == errors.size
+        if version_error_count == errors.size
+          errors.concat check_expected_data_sub_dirs(version_path, version_dir)
+          if version_error_count == errors.size
+            errors.concat check_metadata_dir_files_only(version_path)
+          end
```
ndushay: I'm okay leaving this, but it just occurred to me:  can we now do:

```
version_dir_error_count = errors.size
errors.concat check_sub_dirs(version_sub_dirs, version_dir, EXPECTED_VERSION_SUB_DIRS)
errors.concat check_required_manifest_files(version_path, version_dir) if version_dir_error_count == errors.size
errors.concat check_expected_data_sub_dirs(version_path, version_dir) if version_dir_error_count == errors.size
errors.concat check_metadata_dir_files_only(version_path) if version_dir_error_count == errors.size
```

Is that clearer or less clear?  I'm ok if you want to make a separate refactor ticket for it.

lib/moab/storage_object_validator.rb
```diff
@@ -174,7 +202,7 @@ def check_required_manifest_files(dir, version)
                elsif !has_manifest_inventory && has_signature_catalog
                  result_hash(NO_MANIFEST_INVENTORY, version)
                else
-                 result_hash(NO_XML_FILES, version)
+                 result_hash(NO_FILES_IN_MANIFEST_DIR, version)
```
ndushay: yay!

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +98,36 @@ def check_correctly_formed_moabs
...
+      errors
+    end
+
+    def check_metadata_dir_files_only(version_path)
+      errors = []
+      dir_list = []
+      metadata_dir_path = "#{version_path}/#{DATA_DIR_NAME}/#{METADATA_DIR_NAME}"
+      metadata_sub_dir = sub_dirs(metadata_dir_path)
+      metadata_sub_dir.each { |item| dir_list << File.directory?("#{metadata_dir_path}/#{item}") }
+      errors << result_hash(NO_FILES_IN_METADATA_DIR) if directory_entries(metadata_dir_path).empty?
```
ndushay: yay!

lib/moab/storage_object_validator.rb
```diff
@@ -63,8 +68,13 @@ def version_directories
 
     def check_correctly_named_version_dirs
       errors = []
-      version_directories.each do |version_dir|
-        errors << result_hash(VERSION_DIR_BAD_FORMAT) unless version_dir =~ /^[v]\d{4}$/
+      if version_directories.empty?
+        errors << result_hash(VERSION_DIR_BAD_FORMAT)
```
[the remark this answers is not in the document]
SaravShah: Yup agreed!

lib/moab/storage_object_validator.rb
SaravShah: i like it. makes it more clear!

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +100,70 @@ def check_correctly_formed_moabs
...
+      data_dir_path = "#{version_path}/#{DATA_DIR}"
+      data_sub_dirs = sub_dirs(data_dir_path)
+      if data_sub_dirs.size > EXPECTED_DATA_SUB_DIRS.size
+        errors.concat found_unexpected(data_sub_dirs, version, EXPECTED_DATA_SUB_DIRS)
+      else
+        errors.concat missing_dir(data_sub_dirs, version, [METADATA_DIR]) unless data_sub_dirs.include?(METADATA_DIR)
+        errors.concat found_unexpected(data_sub_dirs, version, EXPECTED_DATA_SUB_DIRS) unless subset?(data_sub_dirs,
+                                                                                                      EXPECTED_DATA_SUB_DIRS)
+
+        if data_sub_dirs.include?('content')
```
ndushay: you should either put the metadata_dir_files check in here, or pull the content dir files check out of here -- consistency is appreciated.

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,155 @@
...
+      context 'under data directory' do
+        it 'has unexpected file' do
+          # v0005
+          expect(verification_array[4]).to eq(Moab::StorageObjectValidator::EXTRA_CHILD_DETECTED =>"Unexpected item in path: [\"extra_file.txt\"] Version: v0005")
+        end
+        it 'has missing metadata directory' do
+          # v0006
+          expect(verification_array[5]).to eq(Moab::StorageObjectValidator::MISSING_DIR =>"Missing directory: [\"metadata\"] Version: v0006")
+        end
+        it 'has missing content directory' do
```
ndushay: this spec description is now wrong

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,155 @@
...
+      end
+      it "has extra characters in version directory name" do
+        druid = 'aa000aa0000'
+        druid_path = 'spec/fixtures/bad_root01/bad_moab_storage_trunk/aa/000/aa/0000/aa000aa0000'
+        storage_obj = Moab::StorageObject.new(druid, druid_path)
+        storage_obj_validator = described_class.new(storage_obj)
+        verification_array = storage_obj_validator.validation_errors
+        expect(verification_array).to include(Moab::StorageObjectValidator::VERSION_DIR_BAD_FORMAT => "Version directory name not in 'v00xx' format")
+      end
+      it "incorrect items in path" do
```
ndushay: i don't understand this error message, sadly -- make a separate ticket to clarify this (patch party).  Is it that unexpected files/dirs are found under the xxxxx (version??) directory?  Under the moab directory?

SaravShah: ticket #97 created!

spec/unit_tests/moab/storage_object_validator_spec.rb
SaravShah: 👍

lib/moab/storage_object_validator.rb
SaravShah: 👍

spec/unit_tests/moab/storage_object_validator_spec.rb
jmartin-sul: agreed, a little unclear.  looks from the file structure of the test fixture like it's unexpected files under the version dir:
```
├── bb
│   └── 000
│       └── bb
│           └── 0000
│               └── bb000bb0000
│                   └── v0001
│                       ├── test1.txt
│                       └── test2.txt
```

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -15,104 +15,155 @@
...
-    end
-    context 'under manifest directory' do
-      druid = 'cc000cc0000'
-      druid_path = 'spec/fixtures/bad_root01/bad_moab_storage_trunk/cc/000/cc/0000/cc000cc0000'
-      storage_obj = Moab::StorageObject.new(druid, druid_path)
-      storage_obj_validator = described_class.new(storage_obj)
-      verification_array = storage_obj_validator.validation_errors
-      it 'does not have signatureCatalog.xml' do
-        expect(verification_array).to include(Moab::StorageObjectValidator::NO_SIGNATURE_CATALOG => "Version: v0001 Missing signatureCatalog.xml")
+      it 'calls #check_expected_data_sub_dirs if moab has the expected version sub dirs' do
```
jmartin-sul: typo?  `#check_expected_data_sub_dirs` -> `check_data_sub_dirs`?

spec/unit_tests/stanford/storage_object_validator_spec.rb
```diff
@@ -8,14 +8,19 @@
     let(:verification_array) { storage_obj_validator1.validation_errors }
```
jmartin-sul: it appears that the `let` statements in this outermost `describe` block aren't actually used?  vars of the same name are used later, but they also seem to be redefined later, within the narrower scopes that use them.

lib/moab/storage_object_validator.rb
```diff
@@ -1,14 +1,17 @@
...
   class StorageObjectValidator
 
-    EXPECTED_DATA_SUB_DIRS = ["content", "metadata"].freeze
-    IMPLICIT_DIRS = ['.', '..'].freeze # unlike Find.find, Dir.entries returns these
-    DATA_DIR_NAME = "data".freeze
-    EXPECTED_VERSION_SUB_DIRS = [DATA_DIR_NAME, "manifests"].freeze
+    METADATA_DIR = "metadata".freeze
+    CONTENT_DIR = "content".freeze
+    EXPECTED_DATA_SUB_DIRS = [CONTENT_DIR, METADATA_DIR].freeze
+    IMPLICIT_DIRS = ['.', '..', '.keep'].freeze # unlike Find.find, Dir.entries returns the current/parent dirs
```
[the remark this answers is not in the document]
ndushay: I would favor a separate issue for this.  I agree with John's point, and have no easy solution handy.  An issue with a patch party label.

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +100,73 @@ def check_correctly_formed_moabs
       version_directories.each do |version_dir|
         version_path = "#{storage_obj_path}/#{version_dir}"
         version_sub_dirs = sub_dirs(version_path)
-        before_result_size = errors.size
-        errors.concat check_sub_dirs(version_sub_dirs, version_dir, EXPECTED_VERSION_SUB_DIRS)
-        after_result_size = errors.size
+        version_error_count = errors.size
+        errors.concat check_version_sub_dirs(version_sub_dirs, version_dir)
         # run the following checks if this version dir passes check_sub_dirs, even if some prior version dirs didn't
```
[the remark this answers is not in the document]
ndushay: or perhaps now the comment isn't even necessary?  Does the code "read" clearly without it?

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +100,73 @@ def check_correctly_formed_moabs
...
+    end
+
+    def check_optional_content_dir_files_only(version_path)
+      errors = []
+      dir_list = []
+      content_dir_path = "#{version_path}/#{DATA_DIR}/#{CONTENT_DIR}"
+      content_sub_dir = sub_dirs(content_dir_path)
+      content_sub_dir.each { |item| dir_list << File.directory?("#{content_dir_path}/#{item}") }
+      errors << result_hash(NO_FILES_IN_CONTENT_DIR) if directory_entries(content_dir_path).empty?
+      errors << result_hash(CONTENT_SUB_DIRS_DETECTED) if dir_list.include?(true)
```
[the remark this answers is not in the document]
ndushay: I agree with John on this.  It would be good to tighten this up.

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +100,73 @@ def check_correctly_formed_moabs
...
+    def check_metadata_dir_files_only(version_path)
+      errors = []
+      dir_list = []
+      metadata_dir_path = "#{version_path}/#{DATA_DIR}/#{METADATA_DIR}"
+      metadata_sub_dir = sub_dirs(metadata_dir_path)
+      metadata_sub_dir.each { |item| dir_list << File.directory?("#{metadata_dir_path}/#{item}") }
+      errors << result_hash(NO_FILES_IN_METADATA_DIR) if directory_entries(metadata_dir_path).empty?
+      errors << result_hash(METADATA_SUB_DIRS_DETECTED) if dir_list.include?(true)
       errors
     end
```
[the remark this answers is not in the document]
ndushay: good 👀  -- yes, please remove unused stuff.  I'm surprised rubocop didn't say anything.

ndushay:
John - thanks for that really thorough review!   I agree with your points.  If we can get @SaravShah to weigh in, maybe we can figure out what should happen next.  I'm online mostly for the evening and can be reached via slack ...

lib/moab/storage_object_validator.rb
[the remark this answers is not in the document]
SaravShah: i'll create a ticket for this!

lib/moab/storage_object_validator.rb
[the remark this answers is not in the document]
SaravShah: i'll remove this since its easy

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +100,73 @@ def check_correctly_formed_moabs
...
       end
+      errors
+    end
+
+    def check_optional_content_dir_files_only(version_path)
+      errors = []
+      dir_list = []
+      content_dir_path = "#{version_path}/#{DATA_DIR}/#{CONTENT_DIR}"
+      content_sub_dir = sub_dirs(content_dir_path)
+      content_sub_dir.each { |item| dir_list << File.directory?("#{content_dir_path}/#{item}") }
```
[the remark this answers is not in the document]
SaravShah: fixed - it was easy

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +100,73 @@ def check_correctly_formed_moabs
...
+    def subset?(first_array, second_array)
+      first_array.to_set.subset?(second_array.to_set)
+    end
+
+    def check_metadata_dir_files_only(version_path)
+      errors = []
+      dir_list = []
+      metadata_dir_path = "#{version_path}/#{DATA_DIR}/#{METADATA_DIR}"
+      metadata_sub_dir = sub_dirs(metadata_dir_path)
+      metadata_sub_dir.each { |item| dir_list << File.directory?("#{metadata_dir_path}/#{item}") }
```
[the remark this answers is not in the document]
SaravShah: done - easy

lib/moab/storage_object_validator.rb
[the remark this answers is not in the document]
SaravShah: 👍

lib/moab/storage_object_validator.rb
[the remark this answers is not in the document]
SaravShah: ticket created #99

lib/moab/storage_object_validator.rb
```diff
@@ -88,18 +100,73 @@ def check_correctly_formed_moabs
...
+      errors << result_hash(NO_FILES_IN_METADATA_DIR) if directory_entries(metadata_dir_path).empty?
+      errors << result_hash(METADATA_SUB_DIRS_DETECTED) if dir_list.include?(true)
```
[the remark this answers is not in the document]
SaravShah: ticket created #99

lib/moab/storage_object_validator.rb
```diff
@@ -88,34 +100,75 @@ def check_correctly_formed_moabs
...
+        errors.concat found_unexpected(data_sub_dirs, version, EXPECTED_DATA_SUB_DIRS)
+      else
+        errors.concat missing_dir(data_sub_dirs, version, [METADATA_DIR]) unless data_sub_dirs.include?(METADATA_DIR)
+        errors.concat found_unexpected(data_sub_dirs, version, EXPECTED_DATA_SUB_DIRS) unless subset?(data_sub_dirs,
+                                                                                                      EXPECTED_DATA_SUB_DIRS)
       end
       errors
     end
 
+    def check_optional_content_dirs_files_only(version_path)
```
ndushay: d'oh:  optional_content_dir_files_only

lib/moab/storage_object_validator.rb
```diff
@@ -88,34 +100,75 @@ def check_correctly_formed_moabs
...
+
+    def subset?(first_array, second_array)
+      first_array.to_set.subset?(second_array.to_set)
+    end
+
+    def check_metadata_dir_files_only(version_path)
+      errors = []
+      dir_list = []
+      metadata_dir_path = "#{version_path}/#{DATA_DIR}/#{METADATA_DIR}"
+      metadata_sub_dir = sub_dirs(metadata_dir_path)
```
ndushay: i think John meant "metadata_sub_dirSSSSS" here.

SaravShah:
Appreciate the in-depth review @ndushay and @jmartin-sul

lib/moab/storage_object_validator.rb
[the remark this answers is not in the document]
jmartin-sul: i'd imagine linting non-use is harder in such a dynamic language.  what if i'm calling it by `send(CHECK_SUB_DIRS_METHOD_NAME)` after setting `CHECK_SUB_DIRS_METHOD_NAME=:check_sub_dirs` :trollface:

PR #95, opened 2017-11-15
Account for optional content/ directory

### LINKED ISSUE #89: (MV) Revision to validity check - data/content is an optional directory
The data/content directory is optional since (a) the object may not have content, and (b) a version may have no new content. MOAB omits the directory rather than creating an empty directory.

tingulfsen:
Closes #89 

A few of the test changes may not make sense at first glance - they were changed due to a previous bug that surfaced due to the code changes.

### DIFF
--- a/lib/moab/storage_object_validator.rb
+++ b/lib/moab/storage_object_validator.rb
@@ -1,12 +1,14 @@
 require 'moab'
+require 'set'
 
 module Moab
   # Given a druid path, are the contents actually a well-formed Moab?
   # Shameless green: repetitious code included.
   class StorageObjectValidator
 
-    EXPECTED_DATA_SUB_DIRS = ["content", "metadata"].freeze
-    IMPLICIT_DIRS = ['.', '..'].freeze # unlike Find.find, Dir.entries returns these
+    METADATA_DIR = 'metadata'.freeze
+    EXPECTED_DATA_SUB_DIRS = ["content", METADATA_DIR].freeze
+    IMPLICIT_DIRS = ['.', '..', '.keep'].freeze # unlike Find.find, Dir.entries returns the current/parent dirs
     DATA_DIR_NAME = "data".freeze
     EXPECTED_VERSION_SUB_DIRS = [DATA_DIR_NAME, "manifests"].freeze
     MANIFEST_INVENTORY_PATH = 'manifests/manifestInventory.xml'.freeze
@@ -63,6 +65,7 @@ module Moab
 
     def check_correctly_named_version_dirs
       errors = []
+      errors << result_hash(MISSING_DIR, 'no versions exist') unless version_directories.count > 0
       version_directories.each do |version_dir|
         errors << result_hash(VERSION_DIR_BAD_FORMAT) unless version_dir =~ /^[v]\d{4}$/
       end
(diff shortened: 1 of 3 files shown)


lib/moab/storage_object_validator.rb
```diff
@@ -63,6 +65,7 @@ def version_directories
 
     def check_correctly_named_version_dirs
       errors = []
+      errors << result_hash(MISSING_DIR, 'no versions exist') unless version_directories.count > 0
```
ndushay: 👍

spec/unit_tests/stanford/storage_object_validator_spec.rb
```diff
@@ -8,14 +8,19 @@
...
     it 'returns errors from calling the superclass validation_errors' do
-      expect(verification_array.count).to eq(2)
+      erroneous_druid = 'xx000xx0000'
+      erroneous_druid_path = 'spec/fixtures/bad_root01/bad_moab_storage_trunk/xx/000/xx/0000/xx000xx0000'
+      erroneous_object = Moab::StorageObject.new(erroneous_druid, erroneous_druid_path)
+      erroneous_object_validator = described_class.new(erroneous_object)
+      error_list = erroneous_object_validator.validation_errors
+      expect(error_list.count).to eq(9)
     end
     it 'returns validation error codes when there is file path for druid to directory validation' do
```
ndushay: I know this was already here, but:  this description doesn't read right.

I like the new error (detecting when there are no version directories) but we do need STANFORD::SOV to have a test to show D2D checking fails.

I think the "missing all version directories" should also be a spec, and it should live in Moab::SOV specs

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -50,9 +50,9 @@
         # v0005
         expect(verification_array[5]).to eq(Moab::StorageObjectValidator::NO_XML_FILES =>"Version: v0005 Missing all required metadata files")
       end
-      it 'has missing content directory' do
+      it 'has missing metadata directory' do
```
ndushay: I like that this is tested here, but do we also have tests for when content directory is missing, and when it is present (both should pass, yes?)

lib/moab/storage_object_validator.rb
```diff
@@ -1,12 +1,14 @@
...
 module Moab
   # Given a druid path, are the contents actually a well-formed Moab?
   # Shameless green: repetitious code included.
   class StorageObjectValidator
 
-    EXPECTED_DATA_SUB_DIRS = ["content", "metadata"].freeze
-    IMPLICIT_DIRS = ['.', '..'].freeze # unlike Find.find, Dir.entries returns these
+    METADATA_DIR = 'metadata'.freeze
+    EXPECTED_DATA_SUB_DIRS = ["content", METADATA_DIR].freeze
+    IMPLICIT_DIRS = ['.', '..', '.keep'].freeze # unlike Find.find, Dir.entries returns the current/parent dirs
```
SaravShah: Nice added .keep

lib/moab/storage_object_validator.rb
SaravShah: Oh nice - I added a check for MISSING_DIR in my PR too - i guess we can just resolve them when we merge. Your solution is more elegant than mine! :D

lib/moab/storage_object_validator.rb
```diff
@@ -89,20 +92,49 @@ def check_correctly_formed_moabs
...
+      else
+        errors.concat missing_dir(data_sub_dirs, version, [METADATA_DIR]) unless data_sub_dirs.include?(METADATA_DIR)
+        errors.concat found_unexpected(data_sub_dirs, version, EXPECTED_DATA_SUB_DIRS) unless subset?(data_sub_dirs,
+                                                                                                      EXPECTED_DATA_SUB_DIRS)
+      end
+      errors
+    end
+
+    def subset?(first_array, second_array)
+      first_array.to_set.subset?(second_array.to_set)
```
SaravShah: 👍

spec/unit_tests/moab/storage_object_validator_spec.rb
SaravShah: +1  for test with missing content dir

spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -50,9 +50,9 @@
...
         # v0006
-        expect(verification_array[6]).to eq(Moab::StorageObjectValidator::MISSING_DIR =>"Missing directory: [\"content\", \"metadata\"] Version: v0006")
+        expect(verification_array[6]).to eq(Moab::StorageObjectValidator::MISSING_DIR =>"Missing directory: [\"metadata\"] Version: v0006")
```
SaravShah: Do we need to have a test that also checks if there are _only_ files in content dir too?

lib/moab/storage_object_validator.rb
ndushay: yeah - I like this solution better.

tingulfsen:
Superseded by #92 - therefore closing.

PR #96, opened 2017-11-15
Use Regexp.match? when possible.

tingulfsen:
This is a performance enhancement:

https://blog.cognitohq.com/new-features-in-ruby-2-4/

The change introduced in this PR gets rid of a few RuboCop warnings that alerted me to this performance improvement.


tingulfsen:
See http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Performance/RegexpMatch also, since you have nothing better to do.

tingulfsen:
We need moab-versioning to run on Ruby pre-2.4 so I am deleting this work.

PR #102, opened 2017-11-22
Use detect

### LINKED ISSUE #99: (MV) storage_object_validator.rb - Refactor to use .detect. (perf improvement?)
line 154: C&P from John
same for line 169

since it appears that the idea here is to see whether any of the content_sub_dir entries are items, you could probably skip the dir_list construction, and just do something like:

```
 errors << result_hash(CONTENT_SUB_DIRS_DETECTED) if content_sub_dirs.detect { |item| File.directory?("#{content_dir_path}/#{item}") } 
```
detect will return the first element for which the block code returns true, or nil if nothing meets the condition
that would feel slightly more readable (and "ruby-ish") to me (and would provide a negligible performance benefit, since it'd stop iterating over the array when it finds the first matching element, though the readability argument is a bigger deal for me).

ndushay:
Small refactoring inspired by #99 (use Enum.detect when checking for subdirs).

- remove misleading private #sub_dirs method, which was effectively an alias for the more aptly named #directory_entries method
- use John's suggestion of Enum.detect when checking for subdirs of `content` and `metadata`, but use a more elegant approach with #contains_sub_dir? 
- remove unnec. TODO
- reword another comment

resolves #99

### DIFF
--- a/lib/moab/storage_object_validator.rb
+++ b/lib/moab/storage_object_validator.rb
@@ -69,7 +69,7 @@ module Moab
     private
 
     def version_directories
-      @vdirs ||= sub_dirs(storage_obj_path)
+      @vdirs ||= directory_entries(storage_obj_path)
     end
 
     def check_correctly_named_version_dirs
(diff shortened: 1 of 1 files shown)


ndushay:
"no tests were changed in the making of this PR"

PR #105, opened 2017-11-27
Allows multiple level sub dir rspec test

SaravShah:
Test shows that we allow multiple sub dir's in data/content/dir, change if we do not want this.


spec/unit_tests/moab/storage_object_validator_spec.rb
```diff
@@ -61,6 +61,15 @@
             # metadata
             expect(verification_array).to include(Moab::StorageObjectValidator::BAD_SUB_DIR_IN_CONTENT_DIR=>"Version v0015: content directory has forbidden sub-directory name: vnnnn or [\"data\", \"manifests\", \"content\", \"metadata\"]")
           end
+          it 'allows multiple level sub directories' do
+            # TODO: if we do not want to allow data/content/dir/dir/dir, remove this test
```
ndushay: actually, this TODO can go -- if we want to forbid multiple level sub directories, we would change this test to "forbids" and flip the expectation.

PR #122, opened 2018-03-06
deposit_bag_validator class added

### LINKED ISSUE #117: (IR) create deposit_bag_validator in MV gem (blocked by PC #601)
given that the moab deposit bag creation code used by "sdr-ingest-tranfer" common_accessioning robot ultimately resides in the Bagger class in moab-versioning, it makes sense for DepositBagValidation code to also live in the Moab gem.   

There also seems to be some overlap between  "sdr-ingest-transfer" robot code in sul-dlss/dor-services/lib/services/sdr_ingest_service  and  "validate-bag" robot code in preservation robots (in sul-dlss/sdr-preservation-core/lib/sdr_ingest/validate_bag).   Possibly more overlap in code that actually puts deposit bag into the moab ...

ndushay:
This is a Shameless Green implementation of Moab::DepositBagValidator, to support the validate_bag robot in sul-dlss/preservation_robots (which will be a replacement for the old robot code in sul-dlss/sdr-preservation-core).

The logic represented by this code was originally in:
- sul-dlss/sdr-preservation-core/lib/sdr_ingest/validate_bag  <-- old preservation robots
- sul-dlss/archive-utils/lib/bagit_bag  <-- gem only used by sdr-preservation-robots
- sul-dlss/archive-utils/lib/file_fixity
- sul-dlss/archive-utils/lib/fixity

The code's new home is here because the common_accessioning robots ultimately calls (via dor-services preservable concern ...)  Moab::Bagger in this moab-versioning gem to create the deposit bag.  So it makes sense to have the bag validation code live here as well.

Please check for:
- [ ] code has at least as much actual validation of the bag as the original
- [ ] code makes sense (suggestions for improvements welcome!)
- [ ] tests check what they should
- [ ] test code makes sense (ditto)
- [ ] does it make sense for this code to live in moab-versioning gem?
- [x] world peace

closes #117
connects to sul-dlss/preservation_robots#11


SaravShah:
mvimg_20180306_153103

ndushay:
In case it helps, some vague parallels here:

Methods:
- #validation_errors --> was sdr-preservation-core.validate_bag#validate_bag
- L67 --> sdr-preservation-core.bagit_bag#verify_pathname
- #bag_dir_exists? --> was archive-utils BagitBag.open_bag
- #verify_version --> was sdr-preservation-core.validate_bag#verify_version_number
- (last_)version_id_from_xxx --> was sdr-preservation-core.validate_bag verify_version_id, vmfile_version_id
- #verify_version_in_xml_file --> sdr-preservation-core.validate_bag#verify_version_id
- #verify_tagmanifests --> archive-utils.BagitBag#verify_tagfile_manifests
- #verify_payload_manifests --> archive-utils.BagitBag#verify_payload_manifests
- #checksums_hash_from_manifest_files --> was archive-utils.BagitBag#read_manifest_files
- #generate_tagmanifest_checksums_hash --> archive-utils:   BagitBag generate_tagfile_checksums, Fixity.generate_checksums
- #generate_payload_checksums --> archive-utils: BagitBag generate_payload_checksums
- #generate_checksums_hash --> archive-utils: Fixity.generate_checksums
- #generated_checksums --> archive-utils: Fixity.fixity_from_file
- #digester_hash --> archive-utils Fixity.get_digesters
- #verify_manifest_checksums --> archive-utils  BagitBag verify_manifests and BagitBag manifest_diff
- #checksums_diff_hash --> archive-utils FileFixity diff
- #verify_payload_size --> archive-utils BagitBag verify_payload_size
- #bag_info_payload_size --> archive-utils  BagitBag info_payload_size, BagitBag read_bag_info_text
- #generated_payload_size --> BagitBag bag_payload_size
- #checksum_types_from_manifest_checksums_hash <-- I created this so we could stop generating unnec checksums
- #single_error_hash <-- akin to MV storage_object_validator
- #error_code_msg <-- akin to MV storage_object_validator

lib/moab/deposit_bag_validator.rb
```diff
@@ -0,0 +1,324 @@
...
+    def initialize(storage_object)
+      @deposit_bag_pathname = storage_object.deposit_bag_pathname
+      @expected_new_version = storage_object.current_version_id + 1
+      @result_array = []
+    end
+
+    # returns Array of tiny error hashes, allowing multiple occurrences of a single error code
+    def validation_errors
+      return [single_error_hash(BAG_DIR_NOT_FOUND, bag_dir: deposit_bag_pathname)] unless deposit_bag_pathname.exist?
+      return result_array unless required_bag_files_exist?
```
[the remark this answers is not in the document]
ndushay: your question points out that that method is doing two things:  writing errors and answering a boolean.  perhaps that is bad.

lib/moab/deposit_bag_validator.rb
```diff
@@ -0,0 +1,324 @@
...
+        when :sha512
+          digesters[checksum_type] = Digest::SHA2.new(512)
+        else
+          result_array << single_error_hash(CHECKSUM_TYPE_UNRECOGNIZED, checksum_type: checksum_type, filename: nil)
+        end
+        digester_hash
+      end
+    end
+
+    def verify_manifest_checksums(manifest_type, manifests_checksum_hash, generated_checksum_hash)
```
SaravShah: i see ``` verify_manifest_checksums``` is the combination of ```verify_manifests``` and ```manifest_diff```. I guess ```manifest_diff``` uses ``` FileFixity.new(file_id: file_id)``` and we are using an empty hash here. Just wanted some clarification

SaravShah:
```     def initialize(options=nil)
      @checksums=Hash.new
      options = {} if options.nil?
      options.each do |key,value|
        #instance_variable_set("@#{key}", value)
        send "#{key}=", value
      end
    end
```

So they are passing in ```file_id``` to this hash obj but we are leaving it empty?

ndushay: So in the original, they use this "FileFixity" structure as a mean of associating a file name with its checksums.  And there is a bunch of code about making FileFixity a class that can be used as a hash key ... but it never is.

When I unraveled what was going on, it appeared we needed something simpler:  a hash with the relative file name as the key, and a hash of checksums as the value.   No need for all the indirection.

We don't have to worry about file name clashes, because they can't happen with relative file names in the bag.  And we are only verifying the checksums that exist for the file, so it seemed expedient to simply work with those.

It may be that some comments would be helpful explaining the hashes.

It may also be that I am wrong in my assessment.

ndushay: (Perhaps it was a bad decision that the original ingest-robots didn't leverage the moab-versioning code for the bag validation, since they do leverage it for deposit-bag.  I have no idea why the existing production code is so inconsistent and strange.)

spec/unit_tests/moab/deposit_bag_validator_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        expect(vdn_errors).not_to include(exp_code => "#{err_msg_prefix}.*\/versionInventory.xml")
+      end
+    end
+
+    it 'result array has CHECKSUM_MISMATCH result hash when tagmanifest checksum verification fails' do
+      bag_pathname = Pathname(File.join(deposit_dir_pathname, 'tagmanifest-checksum-mismatch'))
+      storage_obj = instance_double(Moab::StorageObject, deposit_bag_pathname: bag_pathname, current_version_id: 0)
+      dbv = described_class.new(storage_obj)
+      code = described_class::CHECKSUM_MISMATCH
+      from_file = '6666666666666666666666666666666666666666666666666666666666666666'
```
SaravShah: 👍

lib/moab/deposit_bag_validator.rb
```diff
@@ -0,0 +1,324 @@
...
+    def bag_dir_exists?
+      deposit_bag_pathname.exist?
+    end
+
+    def required_bag_files_exist?
+      REQUIRED_BAG_FILES.each do |filename|
+        pathname = deposit_bag_pathname.join(filename)
+        result_array << single_error_hash(REQUIRED_FILE_NOT_FOUND, file_pathname: pathname) unless pathname.exist?
+      end
+      result_array.empty? ? true : false
```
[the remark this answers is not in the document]
ndushay: adding comment - good catch.

ndushay:
@jmartin-sul I believe I have addressed your concerns.  Thanks for the 👀 .   The code is better with the small tweaks.

lib/moab/deposit_bag_validator.rb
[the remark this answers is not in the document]
atz: `empty?` already returns `true` or `false`.  No need to reinterpret that, right?

PR #136, opened 2018-04-03
Reduce indirection

atz:
Subsume two single-line single-use methods.

Utilize guard clauses and the return value of `.concat` (and other methods).

### DIFF
--- a/lib/moab/storage_object_validator.rb
+++ b/lib/moab/storage_object_validator.rb
@@ -89,15 +89,11 @@ module Moab
 
     # call only if the version directories are "correctly named" vdddd
     def check_sequential_version_dirs
-      errors = []
       version_directories.each_with_index do |dir_name, index|
-        expected_vers_num = index + 1 # version numbering starts at 1, array indexing starts at 0
-        if dir_name[1..-1].to_i != expected_vers_num
-          errors << result_hash(VERSIONS_NOT_IN_ORDER, version_directories)
-          break
-        end
+        next if dir_name[1..-1].to_i == index + 1 # version numbering starts at 1, array indexing at 0
+        return [result_hash(VERSIONS_NOT_IN_ORDER, version_directories)]
       end
-      errors
+      []
     end
 
     def check_correctly_formed_moab(allow_content_subdirs = true)
@@ -113,17 +109,12 @@ module Moab
     end
 
     def check_version_sub_dirs(version_path, version)
-      errors = []
       version_sub_dirs = directory_entries(version_path)
-      version_sub_dir_count = version_sub_dirs.size
-      if version_sub_dir_count == EXPECTED_VERSION_SUB_DIRS.size
-        errors.concat expected_version_sub_dirs(version_path, version)
-      elsif version_sub_dir_count > EXPECTED_VERSION_SUB_DIRS.size
-        errors.concat found_unexpected(version_sub_dirs, version, EXPECTED_VERSION_SUB_DIRS)
-      elsif version_sub_dir_count < EXPECTED_VERSION_SUB_DIRS.size
-        errors.concat missing_dir(version_sub_dirs, version, EXPECTED_VERSION_SUB_DIRS)
-      end
-      errors
+      count = version_sub_dirs.size
+      return expected_version_sub_dirs(version_path, version) if count == EXPECTED_VERSION_SUB_DIRS.size
+      return found_unexpected(version_sub_dirs, version, EXPECTED_VERSION_SUB_DIRS) if count > EXPECTED_VERSION_SUB_DIRS.size
+      return missing_dir(version_sub_dirs, version, EXPECTED_VERSION_SUB_DIRS) if count < EXPECTED_VERSION_SUB_DIRS.size
+      []
     end
 
     def check_data_directory(version_path, version, allow_content_subdirs = true)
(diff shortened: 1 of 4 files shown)


lib/moab/verification_result.rb
```diff
@@ -15,34 +16,30 @@ class VerificationResult
...
+    # @param entity [#to_s] The name of the entity being verified
     # @param expected [Object] The expected value
     # @param found [Object] The found value
     # @return [VerificationResult] The result of comparing the expected and found values
     def self.verify_value(entity, expected, found)
-      result = VerificationResult.new(entity.to_s)
-      result.verified = (expected == found)
-      result.details = { 'expected' => expected, 'found' => found }
-      result
+      new(entity.to_s, (expected == found), 'expected' => expected, 'found' => found)
```
ndushay: I don't find this clearer ... too many expressions inside parens.  And, in this case, i would find the curly braces easier than not including them.  Maybe
```
new(entity.to_s, expected == found, { 'expected' => expected, 'found' => found})
```

it's clearer there are three args, anyway.

lib/moab/storage_object_validator.rb
```diff
@@ -137,15 +128,11 @@ def check_data_directory(version_path, version, allow_content_subdirs = true)
...
-        errors.concat found_unexpected(data_sub_dirs, version, EXPECTED_DATA_SUB_DIRS)
-      else
-        errors.concat missing_dir(data_sub_dirs, version, [METADATA_DIR]) unless data_sub_dirs.include?(METADATA_DIR)
-        errors.concat found_unexpected(data_sub_dirs, version, EXPECTED_DATA_SUB_DIRS) unless subset?(data_sub_dirs,
-                                                                                                      EXPECTED_DATA_SUB_DIRS)
-      end
-      errors
+      errors.concat missing_dir(data_sub_dirs, version, [METADATA_DIR]) unless data_sub_dirs.include?(METADATA_DIR)
+      return errors if data_sub_dirs.to_set.subset?(EXPECTED_DATA_SUB_DIRS.to_set)
+      errors.concat found_unexpected(data_sub_dirs, version, EXPECTED_DATA_SUB_DIRS)
```
ndushay: i think this reads a little easier with 

```ruby
errors.concat xxx unless/if
errors.concat xxx unless/if
errors
```

than

```ruby
errors.concat xxx unless/if
return errors if 
errors.concat xxx
```

in my opinion, the logic is clearer for the first.

lib/moab/verification_result.rb
```diff
@@ -15,34 +16,30 @@ class VerificationResult
...
       @entity = entity
-      @verified = false
-      @details = nil
+      @verified = verified
+      @details = details
       @subentities = Array.new
     end
 
-    # @param entity [Symbol] The name of the entity being verified
+    # @param entity [#to_s] The name of the entity being verified
```
ndushay: copy pasta fail?  it is a Symbol, not a "#to_s"

atz: No, that is deliberate duck-typing.  It is literally anything that responds to `#to_s`, since that is the only thing we do with it.  In no respects to we actually rely on it being a Symbol (though that is expected).

lib/moab/verification_result.rb
```diff
@@ -55,27 +52,21 @@ def to_json(verbose = false)
...
-      hash['verified'] = @verified
-      if (verbose or @verified == false)
-        hash['details'] = @details ? @details : subentities_to_hash(verbose, level)
-      end
-      if level > 0
-        hash
-      else
-        { @entity => hash }
+      hash = { 'verified' => verified }
+      if verbose || verified == false
```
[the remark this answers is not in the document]
atz: You are misreading how operator binding order works so your resistance is based on misunderstanding.  There is no difference in execution with or without parens around the latter clause (or the whole one).  Demonstration:
```ruby
> def foo; puts "FOO"; nil; end
 => :foo 
> def bar; puts "BAR"; 1; end
 => :bar 
> puts "YES" if foo || bar == 1
FOO
BAR
YES
 => nil 
> puts "YES" if foo || (bar == 1)
FOO
BAR
YES
 => nil 

```

lib/moab/verification_result.rb
atz: This is a common syntax for anything that accepts a hash as the last param.  Could make the details hash a separate line and pass that to remove confusion though.

lib/moab/storage_object_validator.rb
atz: Yeah, I can see that.  Not great but whatever.  My point here was mainly to get rid of the single-line single-use `subset?` method in the class.

lib/moab/verification_result.rb
```diff
@@ -15,34 +16,30 @@ class VerificationResult
...
     # @param expression [Object] The expression that will be evaluated as true or false
     # @param details [Object] optional details that could be reported
     # @return [VerificationResult] The result of evaluating the expression
     def self.verify_truth(entity, expression, details = nil)
-      result = VerificationResult.new(entity.to_s)
       # TODO: add expression.empty?
-      result.verified = !(expression.nil? or (expression == false))
-      result.details = details
-      result
+      new(entity.to_s, !(expression.nil? || (expression == false)), details)
```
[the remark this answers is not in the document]
atz: Sure, if you like.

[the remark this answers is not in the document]
atz: If we agree, then it establishes how `verify_truth` is literally no better than an alias for my version of `new` (that takes args).

atz:
@ndushay tweaked as we discussed

PR #137, opened 2018-04-04
Mo rubocop

atz:
This will be the last of my rubocop barrage for the time being...

`VerificationResult cleanup` is manual and the `GuardClause` commit is manual, because it does not have an `--auto-correct`.  

Will rebase once #136 is merged.  (Or merge this one and close that one, whichever.)


Rakefile
```diff
@@ -28,4 +28,4 @@ task :clean do
   FileUtils.rm_r('spec/fixtures/derivatives') if File.exist?('spec/fixtures/derivatives')
 end
 
-task :default => [:spec, :rubocop]
+task :default => %i[spec rubocop]
```
ndushay: i am not a fan of ruby "%xxxx" syntax -- looking up each % thingy to figure out what it means, etc -- I think it obscures rather than clarifies.  I'm disinclined to accept this.  DLSS has new hires that aren't rubyists, and it may not be a ruby shop forever.  Code with the "precious rubyisms" as i call them will be harder to understand.

lib/moab/bagger.rb
```diff
@@ -63,21 +63,21 @@ def create_bagit_txt()
...
     # @return [NilClass] Delete the bagit files
-    def delete_bag()
+    def delete_bag
       # make sure this looks like a bag before deleting
       if bag_pathname.join('bagit.txt').exist?
         if bag_pathname.join('data').exist?
           bag_pathname.rmtree
         else
-          bag_pathname.children.each { |file| file.delete }
+          bag_pathname.children.each(&:delete)
```
ndushay: similarly not a huge fan of this syntax.

atz:
@ndushay You being disinclined to spend time on review is a rather poor justification to block.  I get that time is limited, I just don't like the idea that you look for the minimum nits to object to, put up a block and then not even complete a review.  

I have stylistic variations I might prefer over the default enforced rules, but regularity of the defaults is better than what came before.  If we want to enforce a different style, we can do that mechanically also.  In particular, the `Lint` rules picked up real code mistakes, even ones that were disabled claiming rubocop was wrong.

I am working in this codebase, and I want it to be the version after this PR and not the version before.  I can split it up, but that seems like mostly an exercise in delay.

atz:
The first commit is already it's own PR here: #136

Rakefile
atz: The motivation for this rule is that it forces every element in the array to have the same type, so a stray colon or missing quotes cannot cause later operations to break.  I agree in this specific case it might be overkill for 2 elements, not a great benefit.  I chose to enforce the rule, not to edit this line.

This codebase already uses the similar `%w` operator in 6 different places, so I can't really see the objection overall.

ndushay:
I don't agree that we need all the changes here or even that it is all useful work to be done at this time.  

I think we will have an easier time discussing our differences F2F.

PR #140, opened 2018-04-07
Selectable checksums and matching tests

atz:
I'm putting in the PR now so that work can proceed against the correct method.  We can deal with the system configurability secondarily.

Note: the initializer is removed here, because it *does nothing*. If all a method does is call `super` with the same args, it is better for it not to exist.

Connects to sul-dlss/preservation_catalog#642

### DIFF
--- a/lib/moab/file_signature.rb
+++ b/lib/moab/file_signature.rb
@@ -44,11 +44,6 @@ module Moab
     # The name of the XML element used to serialize this objects data
     tag 'fileSignature'
 
-    # (see Serializable#initialize)
-    def initialize(opts = {})
-      super(opts)
-    end
-
     # @attribute
     # @return [Integer] The size of the file in bytes
     attribute :size, Integer, :on_save => Proc.new { |n| n.to_s }
@@ -64,6 +59,30 @@ module Moab
     # @attribute
     # @return [String] The SHA256 checksum value of the file
     attribute :sha256, String, :on_save => Proc.new { |n| n.nil? ? "" : n.to_s }
+
+    KNOWN_ALGOS = {
+      md5: proc { Digest::MD5.new },
+      sha1: proc { Digest::SHA1.new },
+      sha256: proc { Digest::SHA2.new(256) }
+    }.freeze
+
+    # Reads the file once for ALL (requested) algorithms, not once per.
+    # @param [Pathname] pathname
+    # @param [Array<Symbol>] one or more keys of KNOWN_ALGOS to be computed
+    # @return [Moab::FileSignature] object populated with (requested) checksums
+    def self.from_file(pathname, algos_to_use = KNOWN_ALGOS.keys)
+      raise 'Unrecognized algorithm requested' unless algos_to_use.all? { |a| KNOWN_ALGOS.include?(a) }
+
+      signatures = algos_to_use.map { |k| [k, KNOWN_ALGOS[k].call] }.to_h
+
+      pathname.open("r") do |stream|
+        while (buffer = stream.read(8192))
+          signatures.each_value { |digest| digest.update(buffer) }
+        end
+      end
+
+      new(signatures.map { |k, digest| [k, digest.hexdigest] }.to_h.merge(size: pathname.size))
+    end
 
     # @param type [Symbol,String] The type of checksum
     # @param value [String] The checksum value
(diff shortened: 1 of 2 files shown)


jmartin-sul:
i shouldn't merge since i paired on this, but LGTM.  @SaravShah @ndushay @tallenaz

lib/moab/file_signature.rb
```diff
@@ -44,11 +44,6 @@ class FileSignature < Serializer::Serializable
     # The name of the XML element used to serialize this objects data
     tag 'fileSignature'
 
-    # (see Serializable#initialize)
-    def initialize(opts = {})
-      super(opts)
-    end
```
ndushay: cool!  this is silly cruft; happy to see it go.

lib/moab/file_signature.rb
```diff
@@ -65,6 +60,30 @@ def initialize(opts = {})
     # @return [String] The SHA256 checksum value of the file
     attribute :sha256, String, :on_save => Proc.new { |n| n.nil? ? "" : n.to_s }
 
+    KNOWN_ALGOS = {
+      md5: proc { Digest::MD5.new },
+      sha1: proc { Digest::SHA1.new },
+      sha256: proc { Digest::SHA2.new(256) }
+    }.freeze
```
ndushay: Thrilled to see this pulled out as a constant.   What do you think about making a separate, non-critical-path ticket for it to live in https://github.com/sul-dlss/moab-versioning/blob/master/lib/moab.rb  and also ensuring the constant is used not only here but anywhere else checksums are computed in the code?

lib/moab/file_signature.rb
```diff
@@ -65,6 +60,30 @@ def initialize(opts = {})
...
+    # @param [Pathname] pathname
+    # @param [Array<Symbol>] one or more keys of KNOWN_ALGOS to be computed
+    # @return [Moab::FileSignature] object populated with (requested) checksums
+    def self.from_file(pathname, algos_to_use = KNOWN_ALGOS.keys)
+      raise 'Unrecognized algorithm requested' unless algos_to_use.all? { |a| KNOWN_ALGOS.include?(a) }
+
+      signatures = algos_to_use.map { |k| [k, KNOWN_ALGOS[k].call] }.to_h
+
+      pathname.open("r") do |stream|
+        while (buffer = stream.read(8192))
```
ndushay: in a perfect world, would we also want to call out 8192 as a constant and give it the same treatment?  (is it used anywhere else?).   Again, i would suggest creating a ticket here and consider it non-critical path ...

lib/moab/file_signature.rb
```diff
@@ -139,24 +158,17 @@ def hash
...
-          md5_digest.update(buffer)
-          sha1_digest.update(buffer)
-          sha256_digest.update(buffer)
-        end
-      end
-      @md5 = md5_digest.hexdigest
-      @sha1 = sha1_digest.hexdigest
-      @sha256 = sha256_digest.hexdigest
+      file_signature = self.class.from_file(pathname)
+      @size = file_signature.size
```
ndushay: if these are attributes, should it be self.size???

lib/moab/file_signature.rb
atz: Yes, I would support that.  I think that can be part of the follow-up around configurability, since the configurability will effectively encapsulate the constant.

lib/moab/file_signature.rb
atz: Maybe, but this is a feature in search of use case right now.  Until we have a second value we want to use, I wouldn't touch it.

jmartin-sul: IIRC, @tingulfsen did some testing around this and found that it didn't make a difference once you got past a certain size, but i think a config param would be a totally reasonable backlog ticket.

lib/moab/file_signature.rb
atz: The old instance method is very nearly a constructor.  It sets all 4 attributes and returns the object `self`.  That means that the returned object has basically no relationship to the previous instance, suggesting it should be returning a new object, which is why we deprecated this method and made the class method `self.from_file`.  

For backwards-compatibility, we maintain the method signature and behavior of the old method, but since we stole its guts up in `from_file`, to avoid code duplication, `signature_from_file` now actually does create a new object and takes the attributes from it.  This ensures that we don't have any logical divergence between the two.

Also, more simply, `@size = self.size` is a tautology.

lib/moab/file_signature.rb
atz: @ndushay Already made, referencing your comment

lib/moab/file_signature.rb
atz: I corrected a bunch of those in my separate PR such that if it were merged and this one rebased, I'd expect it to fall out.  The instance variable assignment is obviously from the original implementation.

atz: @ndushay incorporating those changes here.

lib/moab/file_signature.rb
ndushay: cool.  I made #143 and mentioned it in #142.  when the work is done, we'll deal w the dup ticket.

lib/moab/file_signature.rb
ndushay: #144 created.

PR #149, opened 2018-04-11
Don't require pry-byebug on CI

atz:
Obviously we don't debug on Travis and this requirement actually breaks builds at install-time, like:
   https://travis-ci.org/sul-dlss/moab-versioning/jobs/364883575#L585

Error is: `byebug requires Ruby version >= 2.2.0`

Ideally we can eventually drop ruby 2.2 support, but this is a reasonable measure anyway.


ndushay:
Cool!  thanks for fix.

PR #151, opened 2018-04-14
Manually correct GuardClause citations from rubocop

atz:
Also a few lines of indirection removed.

Split out from #137


atz:
@ndushay rebased

PR #152, opened 2018-04-14
Pure autocorrect

atz:
Cherry-picked from #137.  This should be easy to digest as (almost) entirely a bunch of repetitive changes.  With the exception of one file (resolving conflicts), all the code changes were done mechanically via rubocop based on `.rubocop_todo.yml`.  

If a particular style enforced by the tool is objectionable, then IMO, the proper reaction is to add exceptions or configuration to `.rubocop.yml`.  It is counterproductive to leave "offenses" listed in a `TODO` file documented as: `The point is for the user to remove these configuration records one by one as the offenses are removed from the code base.`


atz:
If this gets merged, I'll probably close #137 since I drifted off into trying to rework test that were previously insufficiently differentiated.

lib/moab/file_signature.rb
```diff
@@ -128,12 +128,12 @@ def fixity
...
     def eql?(other)
-      return false unless (other.respond_to?(:size) && other.respond_to?(:checksums))
-      return false if self.size.to_i != other.size.to_i
-      self_checksums = self.checksums
+      return false unless other.respond_to?(:size) && other.respond_to?(:checksums)
+      return false if size.to_i != other.size.to_i
+      self_checksums = checksums
       other_checksums = other.checksums
       matching_keys = self_checksums.keys & other_checksums.keys
-      return false if matching_keys.size == 0
```
jmartin-sul: the intent according to this comment:
https://github.com/sul-dlss/moab-versioning/pull/152/commits/27db28377dd5848764f2a1d98d69146784c96890#diff-11a0d7906801d9dea0eccb85667ad811R75

seemed to be to keep `#size == 0` instead of switching to `#empty`, for compatibility's sake?

lib/moab/signature_catalog.rb
```diff
@@ -183,11 +183,11 @@ def version_additions(version_inventory)
       version_inventory.groups.each do |group|
         group_addtions = FileGroup.new(:group_id => group.group_id)
         group.files.each do |file|
-          unless @signature_hash.has_key?(file.signature)
+          unless @signature_hash.key?(file.signature)
             group_addtions.add_file_instance(file.signature, file.instances[0])
           end
         end
-        version_additions.groups << group_addtions if group_addtions.files.size > 0
+        version_additions.groups << group_addtions unless group_addtions.files.empty?
```
jmartin-sul: ditto the compatibility question from the prior comment

lib/stanford/content_inventory.rb
```diff
@@ -124,7 +128,7 @@ def generate_content_metadata(file_group, object_id, version_id)
     # @return [Boolean] True if contentMetadata has essential file attributes, else raise exception
     def validate_content_metadata(content_metadata)
       result = validate_content_metadata_details(content_metadata)
-      raise Moab::InvalidMetadataException, result[0] + " ..." if result.size > 0
+      raise Moab::InvalidMetadataException, result[0] + " ..." unless result.empty?
```
jmartin-sul: ditto `empty?` (and a few more below, won't mark the rest from here out)

lib/moab/file_signature.rb
atz: It only matters when called on `Pathname` objects, which don't have `empty?` on old rubies.  Here, it is called on an Array, and that's fine.

PR #155, opened 2018-05-22
Include offending dir name to pertaining error codes

### LINKED ISSUE #106: (MV) error messages should include offending file name or directory name
For example, the seeding output gives:  

```E, [2017-11-22T04:55:11.113017 #25241] ERROR -- : PreservedObjectHandler(zz926pj2314, 1, 2806359, <Endpoint: {:endpoint_name=>"services-disk07", :endpoint_type_name=>"online_nfs", :endpoint_type_class=>"online", :endpoint_node=>"localhost", :storage_location=>"/services-disk07/sdr2objects", :recovery_cost=>1}>) Invalid moab, validation errors: ["Should only contain files, but directories were present in the content directory"]```

but it could be nice to know the offending directory name.

### PR DESCRIPTION
**Pull request by SaravShah**
An example log output from seeding
```
[2017-11-22T04:55:11.113017 #25241] ERROR -- :  .............. 
Invalid moab, validation errors: ["Should only contain files, but directories were present in the content directory"]
```

So this PR will give us more details on which directory was present and is not supposed to be there.

 - allow `error_code_msg` to take multiple args other than `addl`
- `METADATA_SUB_DIRS_DETECTED` and `CONTENT_SUB_DIRS_DETECTED`changed `addl` to `version` and added a `dir` variable to tell us the offending directory in logs.

closes #106

### DIFF
--- a/lib/moab/storage_object_validator.rb
+++ b/lib/moab/storage_object_validator.rb
@@ -58,11 +58,11 @@ module Moab
           NO_SIGNATURE_CATALOG => "Version %{addl}: Missing signatureCatalog.xml",
           NO_MANIFEST_INVENTORY => "Version %{addl}: Missing manifestInventory.xml",
           NO_FILES_IN_MANIFEST_DIR => "Version %{addl}: No files present in manifest dir",
-          METADATA_SUB_DIRS_DETECTED => "Version %{version}: metadata directory should only contain files, not directories: %{dir} directory",
+          METADATA_SUB_DIRS_DETECTED => "Version %{version}: metadata directory should only contain files, not directories. Found directory: %{dir}",
           VERSIONS_NOT_IN_ORDER => "Should contain only sequential version directories. Current directories: %{addl}",
           NO_FILES_IN_METADATA_DIR => "Version %{addl}: No files present in metadata dir",
           NO_FILES_IN_CONTENT_DIR => "Version %{addl}: No files present in content dir",
-          CONTENT_SUB_DIRS_DETECTED => "Version %{version}: content directory should only contain files, not directories: %{dir} directory",
+          CONTENT_SUB_DIRS_DETECTED => "Version %{version}: content directory should only contain files, not directories. Found directory: %{dir}",
           BAD_SUB_DIR_IN_CONTENT_DIR => "Version %{addl}: content directory has forbidden sub-directory name: vnnnn or #{FORBIDDEN_CONTENT_SUB_DIRS}"
         }.freeze
     end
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/moab/storage_object_validator.rb [hunk 1]
```diff
@@ -157,7 +158,8 @@ def check_metadata_dir_files_only(version_path)
       errors = []
       metadata_dir_path = File.join(version_path, DATA_DIR, METADATA_DIR)
       errors << result_hash(NO_FILES_IN_METADATA_DIR, basename(version_path)) if directory_entries(metadata_dir_path).empty?
-      errors << result_hash(METADATA_SUB_DIRS_DETECTED, basename(version_path)) if contains_sub_dir?(metadata_dir_path)
+      metadata_sub_dir = contains_sub_dir?(metadata_dir_path)
```
jcoyne: Not a blocker, but I like when methods that end in `?` return a truthy/falsey value, but here you're using the returned value as a string.

lib/moab/storage_object_validator.rb
```diff
@@ -203,12 +205,18 @@ def basename(path)
       path.split(File::SEPARATOR)[-1]
     end
 
-    def result_hash(response_code, addl = nil)
-      { response_code => error_code_msg(response_code, addl) }
+    def result_hash(response_code, msg_args = nil)
+      { response_code => error_code_msg(response_code, msg_args) }
     end
 
     def error_code_msg(response_code, addl = nil)
```
jcoyne: Can you document the parameters here?  It seems like response code is a format string and addl is either a Hash or something, but it's unclear what that case is.

lib/moab/storage_object_validator.rb
```diff
@@ -58,11 +58,11 @@ def self.error_code_to_messages
           NO_SIGNATURE_CATALOG => "Version %{addl}: Missing signatureCatalog.xml",
           NO_MANIFEST_INVENTORY => "Version %{addl}: Missing manifestInventory.xml",
           NO_FILES_IN_MANIFEST_DIR => "Version %{addl}: No files present in manifest dir",
-          METADATA_SUB_DIRS_DETECTED => "Version %{addl}: metadata directory should only contain files, not directories",
+          METADATA_SUB_DIRS_DETECTED => "Version %{version}: metadata directory should only contain files, not directories: %{dir} directory",
           VERSIONS_NOT_IN_ORDER => "Should contain only sequential version directories. Current directories: %{addl}",
           NO_FILES_IN_METADATA_DIR => "Version %{addl}: No files present in metadata dir",
           NO_FILES_IN_CONTENT_DIR => "Version %{addl}: No files present in content dir",
-          CONTENT_SUB_DIRS_DETECTED => "Version %{addl}: content directory should only contain files, not directories",
+          CONTENT_SUB_DIRS_DETECTED => "Version %{version}: content directory should only contain files, not directories: %{dir} directory",
```
jcoyne: These messages would be clearer if they were written as:
```ruby
"...directory should only contain files, not directories. Found directory: '%{dir}'"
```

atz:
@jcoyne: Sarav is out for a bit, so I addressed your issues.

PR #156, opened 2018-05-24
Correct type on two incorrect YARD statements

atz:
Fake YARD!

### DIFF
--- a/lib/moab/storage_repository.rb
+++ b/lib/moab/storage_repository.rb
@@ -56,9 +56,9 @@ module Moab
     end
 
     # @param object_id [String] The identifier of the digital object
-    # @return [Pathname] The branch segment of the object deposit path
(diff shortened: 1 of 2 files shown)


lib/moab/storage_repository.rb
```diff
@@ -56,9 +56,9 @@ def deposit_trunk
     end
 
     # @param object_id [String] The identifier of the digital object
-    # @return [Pathname] The branch segment of the object deposit path
+    # @return [String] The branch segment of the object deposit path
+    # @note Override this method in a subclass
     def deposit_branch(object_id)
-       #todo This method should be customized, or overridden in a subclass
       object_id
```
atz: Should be obvious that the return type has to be the same as the param type, since all we do is return the param.

PR #160, opened 2019-11-15
add InvalidSuriSyntaxError, MoabRuntimeError, MoabStandardError

### LINKED ISSUE #159: 'Identifier has invalid suri syntax' error should have its own class
if it were a subclass of `StandardError`, it'd be easier to trap.  see:
https://github.com/sul-dlss/moab-versioning/blob/6f0759e382647399db5eddc4d94dad5bfeeef400/lib/stanford/storage_repository.rb#L29-L39

for expediency, we've trapped and re-cast this as a more specific error in some pres cat work (see https://github.com/sul-dlss/preservation_catalog/pull/1207).  because it makes the error handling in that controller easier.  but really, that should be happening in this code base.

i don't _think_ this requires a major version bump, because anything that was previously trapping `StandardError` (which is what is raised now) would also be able to trap a subclass of `StandardError` without changing the code.  and if the error message is retained, then any examination of said message would still work.

ndushay:
## Why was this change made?

Closes #159

To fix a kludge in preservation_catalog (see https://github.com/sul-dlss/preservation_catalog/pull/1207/files) that would become more onerous as prescat gets more API functionality to replace sdr-services-app.

## Was the usage documentation (e.g. wiki, README) updated?

N/A

### DIFF
--- a/lib/moab/bagger.rb
+++ b/lib/moab/bagger.rb
@@ -259,7 +259,7 @@ module Moab
       rescue
         shell_execute(tar_cmd.sub('--force-local', ''))
       end
-      raise "Unable to create tarfile #{tar_pathname}" unless tar_pathname.exist?
+      raise(MoabRuntimeError, "Unable to create tarfile #{tar_pathname}") unless tar_pathname.exist?
       true
     end
 
@@ -274,11 +274,11 @@ module Moab
       else
         msg = "Shell command failed: [#{command}] caused by <STDERR = #{stderr}>"
         msg << " STDOUT = #{stdout}" if stdout && stdout.length.positive?
-        raise(StandardError, msg)
+        raise(MoabStandardError, msg)
       end
     rescue SystemCallError => e
       msg = "Shell command failed: [#{command}] caused by #{e.inspect}"
-      raise(StandardError, msg)
+      raise(MoabStandardError, msg)
     end
   end
 end
--- a/lib/moab/exceptions.rb
+++ b/lib/moab/exceptions.rb
@@ -1,13 +1,9 @@
 module Moab
-  class ObjectNotFoundException < RuntimeError
-  end
+  class MoabRuntimeError < RuntimeError; end
+  class MoabStandardError < StandardError; end
 
-  class FileNotFoundException < RuntimeError
-  end
-
-  class InvalidMetadataException < RuntimeError
-  end
-
-  class ValidationException < RuntimeError
-  end
+  class FileNotFoundException < MoabRuntimeError; end
+  class InvalidMetadataException < MoabRuntimeError; end
+  class InvalidSuriSyntaxError < MoabRuntimeError; end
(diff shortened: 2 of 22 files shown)


lib/moab/bagger.rb
```diff
@@ -259,7 +259,7 @@ def create_tarfile(tar_pathname = nil)
       rescue
         shell_execute(tar_cmd.sub('--force-local', ''))
       end
-      raise "Unable to create tarfile #{tar_pathname}" unless tar_pathname.exist?
+      raise(MoabRuntimeError, "Unable to create tarfile #{tar_pathname}") unless tar_pathname.exist?
```
mjgiarlo: This one is trivial: why the parens? I don't think we need them, and it'd look a tad bit more idiomatic to my eyes without them (like before). Totally a style preference, so this isn't blocking; I'm just curious why you added parens to each of these.

lib/moab/exceptions.rb
```diff
@@ -1,13 +1,9 @@
 module Moab
-  class ObjectNotFoundException < RuntimeError
-  end
+  class MoabRuntimeError < RuntimeError; end
```
mjgiarlo: I wonder whether you're open to declaring a single `MoabError` class rather than two classes. Or maybe there is some value to distinguishing between a Moab "standard" error and a Moab "runtime" error? That was there is a single class for all custom errors in this codebase.

lib/moab/bagger.rb
ndushay: I believed it was easier to read with parens when followed by `unless` or `if`.  Perhaps I overdid it.

lib/moab/exceptions.rb
ndushay: The value is that one of the exceptions that used to be thrown is StandardError, so it seemed useful to do it this way.  My goal is to do a minor release of MoabVersion so there is absolutely no ill effect to any app using the gem -- I don't want to cause work for any consumers except pres_cat, where John and Peter did a kludge because they couldn't do this PR at the time.

mjgiarlo: @ndushay Hmm, I'm not sure I follow. Are you saying that the value of making subclasses of StandardError and RuntimeError is that downstream code can rescue the `RuntimeError` children without accidentally rescuing the `StandardError` children?

lib/moab/exceptions.rb
```diff
@@ -1,13 +1,9 @@
...
+  class MoabStandardError < StandardError; end
 
-  class FileNotFoundException < RuntimeError
-  end
-
-  class InvalidMetadataException < RuntimeError
-  end
-
-  class ValidationException < RuntimeError
```
ndushay: ValidationException is not used.

lib/moab/exceptions.rb
ndushay: yup - just in case.

PR #162, opened 2019-12-17
Adds method for getting all instances of storage objects

### LINKED ISSUE #161: Find multiple storage objects
Required by https://github.com/sul-dlss/preservation_catalog/issues/1159

For a druid, find instances across all storage roots.

justinlittman:
Not just first.

closes #161

## Why was this change made?
To support https://github.com/sul-dlss/preservation_catalog/issues/1159 (same druids on multiple storage roots).

## Was the usage documentation (e.g. wiki, README) updated?
No.


.rubocop.yml
```diff
@@ -28,6 +28,10 @@ Lint/UselessAssignment:
     - 'spec/unit_tests/moab/file_group_spec.rb' # TODO: determine if we want to keep this exclusion
     - 'spec/unit_tests/moab/file_inventory_spec.rb' # TODO: determine if we want to keep this exclusion
 
+Metrics/AbcSize:
```
jcoyne: Please move this to `.rubocop_todo.yml`

spec/spec_helper.rb
```diff
@@ -49,6 +49,7 @@ def fixture_setup
   @druid = "druid:#{@obj}"
   @data = @fixtures.join("data", @obj)
   @derivatives = @fixtures.join('derivatives')
+  @derivatives2 = @fixtures.join('derivatives2')
```
jcoyne: Please avoid setting instance vars like this, I know this is what has been done in the past, but the idiomatic way to do this in rspec is to use a `let` declaration just in the contexts where it's needed.

spec/unit_tests/moab/storage_repository_spec.rb
```diff
@@ -39,6 +40,24 @@
     end
   end
 
+  context '#search_storage_objects' do
+    it 'new objects will return an empty array' do
```
jcoyne: The context here is `new objects` and the `it` is `will return empty array`

PR #175, opened 2021-05-20
Update Readme.md to expand examples

julianmorley:
This includes a known-good script that will completely validate a given Druid in Stanford's environment.

## Why was this change made?

The existing example in Readme.md doesn't fully illustrate how to check a Moab in Stanford's environment.

## How was this change tested?

Checked against both known-good and broken druids.

## Which documentation and/or configurations were updated?

Readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,5 +250,3 @@ depending on whether XML serialization is required.
 
 * **Stanford::DorMetadata** = utility methods for interfacing with Stanford metadata files (esp contentMetadata)
   * **Stanford::ActiveFedoraObject** [1..*] = utility for extracting content or other information from a Fedora Instance
-
-
--- a/README.md
+++ b/README.md
@@ -214,8 +214,6 @@ moab.version_list.each do |ver|
 end
 ```
 
-
-
 ## API Documentation
 
 http://rubydoc.info/github/sul-dlss/moab-versioning/main/frames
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,6 +161,64 @@ else
...
+# add to_hash(verbose: true) for more details on each
+  puts ver.verify_version_storage.to_hash
+  puts ver.verify_manifest_inventory.to_hash
+  puts ver.verify_version_inventory.to_hash
+  puts ver.verify_signature_catalog.to_hash
+end
+```
+
+
+
```
ndushay, line 221: can we get rid of some of these extra lines?

README.md
```diff
@@ -195,3 +253,5 @@ depending on whether XML serialization is required.
 
 * **Stanford::DorMetadata** = utility methods for interfacing with Stanford metadata files (esp contentMetadata)
   * **Stanford::ActiveFedoraObject** [1..*] = utility for extracting content or other information from a Fedora Instance
+
+
```
ndushay, line 257: extra lines here as well.

ndushay (commented):
Thank you!  Would you mind terribly removing the extra blank lines?  I can do it in a separate PR if that's a pain for you.

README.md
julianmorley: Done!

ndushay (approved):

PR #183, opened 2021-10-18
[MAJOR] Remove unused method version_metadata

jcoyne:
## Why was this change made?
This was tightly coupled to Fedora 3 and is not used anywhere.


jcoyne:
@ndushay yes, that's right.

PR #192, opened 2022-06-07
[Hold] Fix SimpleCov startup to support Ruby 3.x

aaron-collier:

There is an issue in SimpleCov for ruby >= 3.x with duplicate startup (https://github.com/simplecov-ruby/simplecov/issues/1003). This guards against the duplicate start.

Follow up work will/should include cleaning up deprecations and new cops before pushing a new gem version.

Manually against ruby 2.7.5 and 3.1.1

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -5,9 +5,9 @@ Coveralls.wear_merged! # because we run travis on multiple rubies
 
 require 'simplecov'
 SimpleCov.formatter = Coveralls::SimpleCov::Formatter
-SimpleCov.start do
-  add_filter 'spec'
-end
+SimpleCov.add_filter 'spec'
+
+SimpleCov.start 'rails' unless SimpleCov.running
 
 require 'equivalent-xml'
 require 'moab/stanford'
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -5,9 +5,9 @@
 
 require 'simplecov'
 SimpleCov.formatter = Coveralls::SimpleCov::Formatter
-SimpleCov.start do
-  add_filter 'spec'
-end
+SimpleCov.add_filter 'spec'
+
+SimpleCov.start 'rails' unless SimpleCov.running
```
ndushay, line 10: but ... this gem isn't rails ...

See also https://github.com/sul-dlss/moab-versioning/pull/193.

I'm fine with this to get it going ...

ndushay (approved):
for your consideration - this gem isn't rails, so the SimpleCov.start statement seems wrong to me but I don't grok it ...

ndushay:
this PR/branch will be combined with #193 in a new PR.

aaron-collier:
Closing in favor of #193

Issue #194, opened 2022-06-08
Refactor testing for reliability and updated patterns


aaron-collier:
The test suite is significantly out of date and introduces several code smells that should be given attention to refactor. One result of this is tests that fail when run individually.

For instance, the full suite passes:

```
moab-versioning on  main [!] via 💎 v3.1.1 on ☁️  (us-west-2) took 6m22s 
❯ bundle exec rspec
.
.
.
Finished in 3.8 seconds (files took 0.38009 seconds to load)
403 examples, 0 failures

Coverage report generated for RSpec to /Users/amcollie/github/sul-dlss/moab-versioning/coverage. 1654 / 1691 LOC (97.81%) covered.
```

However, running the `storage_object_spec` alone fails:

```
❯ bundle exec rspec spec/unit_tests/moab/storage_object_spec.rb
.............F...WARNING: Using `expect { }.not_to raise_error(SpecificErrorClass)` risks false positives, since literally any other error would cause the expectation to pass, including those raised by Ruby (e.g. `NoMethodError`, `NameError` and `ArgumentError`), meaning the code you are intending to test may not even get reached. Instead consider using `expect { }.not_to raise_error` or `expect { }.to raise_error(DifferentSpecificErrorClass)`. This message can be suppressed by setting: `RSpec::Expectations.configuration.on_potential_false_positives = :nothing`. Called from /Users/amcollie/github/sul-dlss/moab-versioning/spec/unit_tests/moab/storage_object_spec.rb:362:in `block (3 levels) in <top (required)>'.
.........

Failures:

  1) Moab::StorageObject #size
     Failure/Error:
       Find.find(object_pathname) do |path|
         size += FileTest.size(path) unless FileTest.directory?(path)
       end
     
     NameError:
       uninitialized constant Moab::StorageObject::Find
     
             Find.find(object_pathname) do |path|
             ^^^^
     # ./lib/moab/storage_object.rb:230:in `size'
     # ./spec/unit_tests/moab/storage_object_spec.rb:474:in `block (2 levels) in <top (required)>'

Finished in 0.32443 seconds (files took 0.26354 seconds to load)
26 examples, 1 failure

Failed examples:

rspec ./spec/unit_tests/moab/storage_object_spec.rb:470 # Moab::StorageObject #size

Coverage report generated for RSpec to /Users/amcollie/github/sul-dlss/moab-versioning/coverage. 1182 / 1691 LOC (69.9%) covered.
```

[...]

PR #195, opened 2022-06-08
Fix minor issue/difference between circle tests and local

aaron-collier:

There is a small difference between the data output on circle and locally, so this line is required to allow the tests to pass locally and in circle (line 156).

This highlights a smell that should be addressed as part of #194 


ndushay (changes requested):
I think it would be better to just turn off, or increase the ExampleLength in `.rubocop.yml`  Because to hell with that -- these are specs, for goodness sake.

aaron-collier:
@ndushay works for me :-). I've excludes specs for that cop.

ndushay (approved):

PR #197, opened 2022-06-09
comment about thread safety on class var in `StorageServices` classes, additional related test

### LINKED ISSUE #189: Make moab-versioning thread safe
from @jcoyne in Slack thread

> The only problem I see in moab-versioning around threading is https://github.com/sul-dlss/moab-versioning/blob/main/lib/moab/storage_services.rb#L17 and it would be pretty trivial to switch that to a singleton.
>
> Then move some of the lazy config in StorageRepository to the initializer and Bob’s your uncle.

image

aaron-collier:

Fixes #189 

This wraps the class variable in a class method for easier readability and refactoring, and tests that the respective `StorageRepository` for each is only instantiated once.

### DIFF
--- a/lib/moab/storage_services.rb
+++ b/lib/moab/storage_services.rb
@@ -13,81 +13,90 @@ module Moab
   # @note Copyright (c) 2012 by The Board of Trustees of the Leland Stanford Junior University.
   #   All rights reserved.  See {file:LICENSE.rdoc} for details.
   class StorageServices
-    # @return [StorageRepository] an instance of the interface to SDR storage
+    # @note After some discussion, consensus was that this is a thread safe use of a
+    # class variable, as 1) it's never mutated after the class is initialized, and 2) the
+    # value of the StorageRepository instance is determined from configuration that
+    # rarely changes and is loaded once at app start time (at least in Stanford's
+    # consumers; see Moab::Config.configure calls in preservation_robots, preservation_catalog,
+    # and technical-metadata-service).
+    # Sidekiq requires thread safe code, so please preserve thread safety for multiple
+    # concurrent callers of this service if refactoring, so Sidekiq remains an option for
+    # ActiveJob backend for moab-versioning consumers.
     @@repository = Moab::StorageRepository.new
 
+    # @return [StorageRepository] an instance of the interface to SDR storage
     def self.repository
       @@repository
     end
 
     # @return [Array<Pathname>] A list of the filesystems currently used for storage
     def self.storage_roots
-      @@repository.storage_roots
+      repository.storage_roots
     end
 
     # @return [String] The trunk segment of the object deposit path
     def self.deposit_trunk
-      @@repository.deposit_trunk
+      repository.deposit_trunk
     end
 
     # @param object_id [String] The identifier of the digital object
     # @return [Pathname] The branch segment of the object deposit path
     def self.deposit_branch(object_id)
-      @@repository.deposit_branch(object_id)
+      repository.deposit_branch(object_id)
     end
 
     # @param object_id [String] The identifier of the digital object
     # @param [Object] include_deposit
     # @return [StorageObject] The representation of a digitial object's  storage directory, which might not exist yet.
     def self.find_storage_object(object_id, include_deposit = false)
-      @@repository.find_storage_object(object_id, include_deposit)
+      repository.find_storage_object(object_id, include_deposit)
     end
 
     # @param object_id [String] The identifier of the digital object
     # @param [Object] include_deposit
     # @return [Array<StorageObject>] Representations of a digitial object's storage directories, or an empty array if none found.
     def self.search_storage_objects(object_id, include_deposit = false)
-      @@repository.search_storage_objects(object_id, include_deposit)
+      repository.search_storage_objects(object_id, include_deposit)
     end
 
     # @param object_id [String] The identifier of the digital object whose size is desired
     # @param include_deposit [Boolean] specifies whether to look in deposit areas for objects in process of initial ingest
     # @return [Integer] the size occupied on disk by the storage object, in bytes.  this is the entire moab (all versions).
     def self.object_size(object_id, include_deposit = false)
-      @@repository.object_size(object_id, include_deposit)
+      repository.object_size(object_id, include_deposit)
     end
 
     # @param object_id [String] The identifier of the digital object whose version is desired
     # @param create [Boolean] If true, the object home directory should be created if it does not exist
     # @return [StorageObject] The representation of a digitial object's storage directory, which must exist.
     def self.storage_object(object_id, create = false)
-      @@repository.storage_object(object_id, create)
+      repository.storage_object(object_id, create)
     end
 
     # @param object_id [String] The digital object identifier of the object
     # @return [String] the location of the storage object
     def self.object_path(object_id)
-      @@repository.storage_object(object_id).object_pathname.to_s
+      repository.storage_object(object_id).object_pathname.to_s
     end
 
     # @param object_id [String] The digital object identifier of the object
     # @param [Integer] version_id The ID of the version, if nil use latest version
     # @return [String] the location of the storage object version
     def self.object_version_path(object_id, version_id = nil)
-      @@repository.storage_object(object_id).find_object_version(version_id).version_pathname.to_s
+      repository.storage_object(object_id).find_object_version(version_id).version_pathname.to_s
     end
 
     # @param object_id [String] The digital object identifier
     # @return [Integer] The version number of the currently highest version
     def self.current_version(object_id)
-      @@repository.storage_object(object_id).current_version_id
+      repository.storage_object(object_id).current_version_id
     end
 
     # @param [String] object_id The digital object identifier of the object
     # @param [Integer] version_id The ID of the version, if nil use latest version
     # @return [FileInventory] the file inventory for the specified object version
     def self.retrieve_file_group(file_category, object_id, version_id = nil)
-      storage_object_version = @@repository.storage_object(object_id).find_object_version(version_id)
+      storage_object_version = repository.storage_object(object_id).find_object_version(version_id)
       inventory_type = if file_category =~ /manifest/
                          file_category = 'manifests'
                        else
@@ -103,7 +112,7 @@ module Moab
     # @param [Integer] version_id The ID of the version, if nil use latest version
     # @return [Pathname] Pathname object containing the full path for the specified file
     def self.retrieve_file(file_category, file_id, object_id, version_id = nil)
-      storage_object_version = @@repository.storage_object(object_id).find_object_version(version_id)
+      storage_object_version = repository.storage_object(object_id).find_object_version(version_id)
       storage_object_version.find_filepath(file_category, file_id)
     end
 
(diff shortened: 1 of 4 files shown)


ndushay:
I think we agreed to hold off on untangling this for now, as Stanford flavor repos shouldn't be used for testing the super classes ... but there are apparently a lot of tests with this problem...

If I'm wrong, take it out of draft.

ndushay:
@jmartin-sul since there is already test code that uses Stanford specific code to test the generic classes, I'm inclined to pull this out of draft and merge it.  Do you agree?

Aaron already ticketed untangling the specs in #194

lib/moab/storage_services.rb
```diff
@@ -14,80 +14,79 @@ module Moab
   #   All rights reserved.  See {file:LICENSE.rdoc} for details.
   class StorageServices
     # @return [StorageRepository] an instance of the interface to SDR storage
-    @@repository = Moab::StorageRepository.new
 
     def self.repository
-      @@repository
+      @repository ||= Stanford::StorageRepository.new
```
jmartin-sul, line 19: So... I haven't done a whole new web search to refresh myself on the problem, but when I was looking into Ruby multi-threading issues and Sidekiq/Resque stuff a couple years ago, I found this blog post (one of the things I actually bookmarked on the subject) particularly helpful: https://bearmetal.eu/theden/how-do-i-know-whether-my-rails-app-is-thread-safe-or-not/

Anyway, one of my takeaways from it was that class vars (`@@`) and class instance vars (`@` ivars in a class method) are equally dangerous, and similarly subject to multi-threading issues.

Re-reading that article today, I have a few thoughts:
1. The most dangerous pitfalls in this sort of multi-threading situation seem to be around mutations where a calculation might overwrite shared state in such a way that a stale value is used in a calculation that then overwrites a fresher value.  The simple but common pitfall they point out is something like a shared class var counter incremented by the `+=` operator, which is actually two operations... but as they observe, so is the `||=`.  But in the case of our code here, the possible overwrite will almost certainly just overwrite the existing value with the same value, because the `Stanford::StorageRepository` instance will be based on the same configuration each time (at least the way we use moab-versioning -- maybe this won't be the case for other consumers? :shrug:).  Which is to say, maybe this is technically still not thread safe, but also, maybe it doesn't matter?  But this does worry me a bit in terms of finding other thread un-safe operations in the gem, because as that post observed, a class ivar is hard to distinguish from an object ivar, at least just by grepping for `@`.
2. One way out: we could just not memoize this value.  I suspect that we wouldn't experience a noticeable performance hit, because other perservation versioning operations like updating files and checking checksums are relatively way more expensive... but I'm not sure the performance hit would be negligible.  "Premature optimization is the root of all evil" -some famous computer scientist (I think it was Wirth, but am not sure off-hand 😂)
3. Another way out: per that blog post, we could memoize in a thread local var, e.g. `Thread.current[:repository] ||= Stanford::StorageRepository.new`.  That might be my preference: seems like it'd be the least disruptive change in terms of the expected behavior and interface from the consumer's perspective, while also providing the desired thread safety.  As the blog post points out, it sort of smells like a global var, but I don't think it smells anymore like that than what the gem does at present 🙂 
4. I'm not totally sure, but I don't think explicitly splitting the `||=` into separate atomic operations would make a difference (e.g. a `return` guard clause if the class var exists, and memoization if the guard clause falls through).
5. We could make it the responsibility of the consumer to instantiate `Stanford::StorageRepository` locally, but that sounds like a pain to coordinate across the codebases, so I'd love to avoid that approach.

This all makes me sort of want to just grep for memoization in this gem to look for other similar issues... a super cursory `git grep '||='` makes me think that the other memoizations are actually just regular object ivars (as opposed to class instance vars like this).  But not 100% sure about that.

jmartin-sul: TLDR i vote that we tweak this line to be `Thread.current[:repository] ||= Stanford::StorageRepository.new`, and see how that works.

jmartin-sul: one last thing, not really a suggestion, but interested in others' thoughts on it: one thing that the blog post suggests is a `Mutex#synchronize` wrapper around the thread-unsafe code... but the blog post also suggests creating and storing the `Mutex` in a way that seems not thread-safe, because it suggests putting it in a `@@` class var?  and i'm having a hard time thinking of how one would even share the mutex in a safe way, given that.

jmartin-sul (commented):
was torn between "comment" and "request changes".  went with the former, because i think this is an improvement, since it hides `repository` behind a method that should make it easier to refactor.  but i'm also not sure it actually addresses the multi-threading issue.

lib/moab/storage_services.rb
jmartin-sul: nope, wrong attribution, it was knuth (of course)

jmartin-sul: ugh, an now, after thinking about this more and reading about it more, i'm left wondering if the current implementation on `main` is maybe thread-safe afterall... if the class is only instantiated once per process, and the class var is only set at instantiation and never mutated, is there actually a worry around concurrency and atomicity?

maybe we should flag this for post-standup discussion monday.  might help to have multiple sets of eyes on this, esp if some of the viewers better grok the way ruby classes are managed by the interpreter than i do.

aaron-collier: Post standup discussion (or infra dev) is totally warranted here I think. Thanks for the thoughtful comments :-)

jmartin-sul: np!

ndushay:
@jcoyne - we'd love your opinion on this;  we're unclear on our thread-safety risk.

@jmartin-sul will be in charge of shepherding this forward

One suggestion was to add a one-line comment before L19 referring to this PR's comments.

ndushay:
another suggestion was to ask someone more knowledgeable of thread safety in ruby ... someone out in the big bad world.

lib/moab/storage_services.rb
```diff
@@ -14,80 +14,79 @@ module Moab
   #   All rights reserved.  See {file:LICENSE.rdoc} for details.
   class StorageServices
     # @return [StorageRepository] an instance of the interface to SDR storage
```
jcoyne, line 16: I was expecting this to `include Singleton` see: https://ruby-doc.org/stdlib-3.0.0/libdoc/singleton/rdoc/Singleton.html

jcoyne (dismissed):

lib/moab/storage_services.rb
jmartin-sul: ok, i think i made this change as suggested (in the second commit).

i added a test, but i'm not sure it's a worthwhile test?

spec/unit_tests/moab/storage_services_spec.rb
```diff
@@ -3,6 +3,14 @@
 describe Moab::StorageServices do
   let(:eq_xml_opts) { { element_order: false, normalize_whitespace: true } }
 
+  it 'is a singleton' do
+    # see https://ruby-doc.org/stdlib-3.0.0/libdoc/singleton/rdoc/Singleton.html
+    described_class.instance
+    # providing an empty block returns a count, otherwise, an iterator is returned
+    class_instance_count = ObjectSpace.each_object(described_class) {} # rubocop:disable Lint/EmptyBlock
+    expect(class_instance_count).to eq 1
+  end
```
jmartin-sul, line 12: this is the part i wanted to pair on:  this doesn't feel like a very realistic test, and so i'm not sure how worthwhile it is, but i had trouble writing a more realistic test that showed showed that this is a `Singleton`.

ideally, i'd like to test the desired behavior, which is that `@@repository` (now exposed by the `repository` method) is only initialized once.

jmartin-sul (commented):
per standup and storytime discussion
* i've pushed up my WIP changes, and left a comment on the iffy test that i wanted help on.
* i'll mark this WIP and unassign myself and throw it back in ready

spec/unit_tests/moab/storage_services_spec.rb
ndushay: could you instantiate it and then do something else and confirm it's the same object with === ?

jmartin-sul: the usage of the `StorageServices` classes is all via class methods, so i don't know that it really makes sense to instantiate them.  but i did end up adding tests that simulate calls from multiple threads, since that's the use case we're worried about, and those tests expect only one instantiation of `StorageRepository`.  hopefully that suffices?

.rubocop.yml
```diff
@@ -209,3 +209,8 @@ Lint/ConstantOverwrittenInRescue: # new in 1.31
   Enabled: true
 Lint/NonAtomicFileOperation: # new in 1.31
   Enabled: true
+
+RSpec/Capybara/SpecificMatcher: # new in 2.12
+  Enabled: true
+RSpec/Rails/HaveHttpStatus: # new in 2.12
+  Enabled: true
```
jmartin-sul, line 216: no actual code changes resulted from this, so i figured it was ok to throw on one small extra commit to get rid of the warnings, but happy to split out if preferred.

spec/unit_tests/moab/storage_services_spec.rb
```diff
@@ -3,6 +3,21 @@
 describe Moab::StorageServices do
   let(:eq_xml_opts) { { element_order: false, normalize_whitespace: true } }
 
+  it 'only instantiates StorageRepository once for the class, even if multiple threads try to access it' do
+    # Object#object_id Returns an integer identifier for obj.  The same number will be returned
+    # on all calls to object_id for a given object, and no two active objects will share an id.
+    # https://ruby-doc.org/core-3.0.1/Object.html#method-i-object_id
```
jmartin-sul, line 9: perhaps an unnecessary comment, can remove if people feel it's overkill or distraction

jmartin-sul:
i think this is finally ready for re-review

jmartin-sul:
pushed a small change to the tests that make them slightly more concise and slightly more thorough.

ndushay (approved):

.rubocop.yml
jcoyne: I think it's kind of odd to enable Rails cops for this repo, which doesn't have a dependency on Rails, but it doesn't really matter much.

lib/moab/storage_services.rb
```diff
@@ -13,81 +13,86 @@ module Moab
   # @note Copyright (c) 2012 by The Board of Trustees of the Leland Stanford Junior University.
   #   All rights reserved.  See {file:LICENSE.rdoc} for details.
   class StorageServices
-    # @return [StorageRepository] an instance of the interface to SDR storage
+    include Singleton
+
+    # this being a Singleton should result in only one class instance being created for
```
jcoyne, line 18: I'm not sure this follows.   A singleton is about the object instance, not about the class instance. I'm don't believe`include Singleton` is doing anything here.

jcoyne (changes requested):

.rubocop.yml
jmartin-sul: good point, dropped

lib/moab/storage_services.rb
jmartin-sul: yeah, agreed, i ended up thinking the same after discussion on slack week before last and seeing the new test continue to pass even when `include Singleton` was commented out.  pushed up another update removing the `include` and updating the comment to hopefully be more accurate.

spec/unit_tests/moab/storage_services_spec.rb
jmartin-sul: i think i may have a worthwhile test now in this PR?

jcoyne (approved):

PR #202, opened 2022-07-06
Expose the use_links parameter on ingest_bag

### LINKED ISSUE #200: allow Moab::StorageObjectVersion#ingest_dir and #ingest_file to take a caller supplied value for their use_links parameter
These are called by `Moab::StorageObject#ingest_bag` (which is called by pres robots [`Robots::SdrRepo::PreservationIngest::UpdateMoab#update_moab`](https://github.com/sul-dlss/preservation_robots/blob/da609e494bd91511b65649cc4134f74325ff885e/lib/robots/sdr_repo/preservation_ingest/update_moab.rb#L26)).

Making this parameter accessible will allow pres robots to supply a flag that lets it use the existing copy option when creating a new Moab version, instead of the default hardlinking behavior, which we hope will make our preservation system play better with Ceph backed storage on which our preservation content sits.

connects with (and blocks) https://github.com/sul-dlss/preservation_robots/issues/405

see https://github.com/sul-dlss/moab-versioning/blob/1cfe59cfb4f9fea7a1cc35b936a8f5a7ddd94029/lib/moab/storage_object.rb#L62-L78

and https://github.com/sul-dlss/moab-versioning/blob/85911a0b48aeca63a79ae339a86c88c989161fdd/lib/moab/storage_object_version.rb#L141-L180

jcoyne:

Fixes #200

This allows us to do a normal move operation rather than using links.

Test suite

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -16,6 +16,8 @@ GEM
       minitest (>= 5.1)
       tzinfo (~> 2.0)
     ast (2.4.2)
+    byebug (11.1.3)
+    coderay (1.1.3)
     concurrent-ruby (1.1.10)
     deprecation (1.1.0)
       activesupport
@@ -28,7 +30,10 @@ GEM
     i18n (1.10.0)
       concurrent-ruby (~> 1.0)
     json (2.6.2)
+    method_source (1.0.0)
     minitest (5.16.2)
+    nokogiri (1.13.6-x86_64-darwin)
+      racc (~> 1.4)
     nokogiri (1.13.6-x86_64-linux)
       racc (~> 1.4)
     nokogiri-happymapper (0.9.0)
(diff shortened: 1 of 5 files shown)


ndushay (changes requested):
did you mwan for Gemfile.lock to change?  Looks like maybe you were debugging??

Also suggest using `File.symlink?` for the tests

jcoyne:
@ndushay The Gemfile.lock changes happened when I did `bundle install`, I split that into: https://github.com/sul-dlss/moab-versioning/pull/203 to be more tightly scoped, but I didn't add any dependencies.

ndushay (approved):
I'm probably being ignorant here, but wouldn't a "hardlink" show up as a symlink?  And wouldn't we want to check that the result of running the method was different for the diff values of the variable?   Wouldn't `File.symlink?` be true for the hardlink for files, and not true if it's the actual file?

Please educate me if I'm being ignorant.

jcoyne:
Hard links are actually real files. They are a pointer to an inode. They are materially different from soft links.

jmartin-sul (commented):
question about a possible comment tweak to consider for a follow-on PR

PR #208, opened 2022-07-29
Performance improvement in storage_object_version.

justinlittman:
Performance

techmd-prod

It reduced listing files for an item with 421 files from 162 seconds to 1.23 seconds. Verified that the results were the same.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -16,6 +16,8 @@ GEM
       minitest (>= 5.1)
       tzinfo (~> 2.0)
     ast (2.4.2)
+    byebug (11.1.3)
+    coderay (1.1.3)
     concurrent-ruby (1.1.10)
     deprecation (1.1.0)
       activesupport
@@ -28,6 +30,7 @@ GEM
     i18n (1.12.0)
       concurrent-ruby (~> 1.0)
     json (2.6.2)
+    method_source (1.0.0)
     minitest (5.16.2)
     nokogiri (1.13.8-x86_64-darwin)
       racc (~> 1.4)
@@ -38,6 +41,12 @@ GEM
     parallel (1.22.1)
     parser (3.1.2.0)
       ast (~> 2.4.1)
+    pry (0.13.1)
+      coderay (~> 1.1)
+      method_source (~> 1.0)
+    pry-byebug (3.9.0)
+      byebug (~> 11.0)
+      pry (~> 0.13.0)
     racc (1.6.0)
     rainbow (3.1.1)
     rake (13.0.6)
@@ -89,6 +98,7 @@ PLATFORMS
 DEPENDENCIES
   equivalent-xml
   moab-versioning!
+  pry-byebug
   rake
   rspec
   rubocop (~> 1.7)
--- a/lib/moab/storage_object_version.rb
+++ b/lib/moab/storage_object_version.rb
@@ -128,11 +128,11 @@ module Moab
     # @api external
     # @return [SignatureCatalog] The signature catalog of the digital object as of this version
     def signature_catalog
-      if version_id > 0
-        SignatureCatalog.read_xml_file(@version_pathname.join('manifests'))
-      else
-        SignatureCatalog.new(digital_object_id: @storage_object.digital_object_id)
-      end
+      @signature_catalog ||= if version_id > 0
+                               SignatureCatalog.read_xml_file(@version_pathname.join('manifests'))
+                             else
+                               SignatureCatalog.new(digital_object_id: @storage_object.digital_object_id)
+                             end
     end
 
     # @api internal


jcoyne (approved):

ndushay (approved):

ndushay:
From @jmartin-sul in slack:

>based on what i know, i don't think i'd worry much about a situation where multiple processes (e.g. multiple pres robots workers, or a pres robots worker and a pres cat worker) have possibly stale signature catalog objects in memory, since i think the processes that instantiate those probably aren't running concurrently, and probably aren't running for too long (except maybe when pres cat is checksum validating a large moab).  like... pres cat shouldn't be reading a thing when pres robots is still writing it (possible ceph buffering issues notwithstanding).  and only one pres robots worker should be working on a given druid at a time, because WFS?  i'll do some looking at the usage in the gem, but the only worry that jumps out at me is whether moab-versioning itself might mutate the signature catalog, and then act on a stale version of the info due to the memoized object.  but of course i'd hope the test suite would catch an issue like that.
i have no prediction as to how this would play with ceph issues, either in terms of the possibly increased update throughput, or possibly reduced file read load :shrug:

jmartin-sul (approved):
I read through the calls to the updated method, threw in a little debug logging locally to see whether the memoization gets exercised by the tests, and looked through the pres cat and pres robots codebases, and I think this is fine for the following reasons:
* The memoization does get used by the unit tests in this codebase, in particular `storage_object_spec.rb:137`, the `creates versionInventory and versionAdditions` test.  Also the `#verify_version_storage`, `#find_filepath`, `#verify_version_inventory`, and `#verify_signature_catalog`  tests in `storage_object_version_spec.rb`.
* A number of places in this codebase, particularly in `lib/moab/storage_object.rb` and `lib/moab/storage_object_version.rb` already do this sort of memoization to a local var within their own methods, instead of fetching a freshly computed value from the  `#signature_catalog` method each time a signature catalog is needed.
* I hadn't thought this through fully on Friday, since I didn't get to spend much time with this PR till today, but intuitively it makes sense that the file being read by the code this PR updates would never be mutated out from under the code, because the code in question is in `storage_object_version.rb`.  Once written, a given Moab version shouldn't ever be changing anyway.  It would make sense that the XML for a new version's signature catalog would be computed based on a diff from the old version, written, and then never (intentionally) updated.  I don't see any evidence, from looking at the code, that it's written and mutated before being considered "finalized", or something like that.  `StorageObjectVersion#update_catalog` calls `SignatureCatalog#write_xml_file`.  `StorageObjectVersion#update_catalog` is tested (and if i update it to check that `write_xml_file` is only called once, the tests still pass).  `StorageObjectVersion#update_catalog` is called once by `StorageObject#ingest_bag`, which is called once by pres robots when it runs the `update-moab` step for a given druid version.
* preservation_robots, the only code we expect to update Moabs in preservation storage, doesn't directly call this method at all (only indirectly as described above).
* preservation_catalog does call this method directly, but it mounts the pres storage as read-only, and uses the moab-versioning gem for checksum validation.  For checksum validation as part of accessioning (`validate-moab` step in `preservationIngestWF`), we'd expect that nothing is updating the Moab while the validation is running (because nothing else should be able to modify that druid while `accessionWF` is running, and workflow steps happen sequentially with no parallelization, at least as far as is known/intended).  For checksum validation as part of regularly scheduled audits, there's always the slim the chance that the validation will happen to be running when a new version of an existing object is being accessioned, but this PR doesn't change the likelihood of that happening, and we've never seen this sort of clash afaik (and a re-run of checksum validation after things settle would be expected to pass, and it should be obvious from the object's workflow and event history that the audit was running while the object was being versioned).
* Because the `validate-moab` step is there as a safeguard in `preservationIngestWF`, and because nothing should be updating a Moab while that step runs on a new version in `accessionWF` (per the above explanation), we should have an essentially unchanged checksum validation step in place that should catch whatever wonky/inconsistent state an object might end up in if there was some ill effect from this change that we missed.  It seems very unlikely that some concurrency issue during the `update-moab` step of `preservationIngestWF` would lead to a stale signature catalog which would be used in a comparison that'd still result in a (seemingly) valid and internally consistent new Moab version being written that was nonetheless missing a piece of checksum data, or which contained extraneous or incorrect checksum data.

If we knew of a situation where we intentionally allowed multiple processes to create new Moab versions simultaneously, or if we intentionally modified existing Moab versions once written, I might not think the above analysis sufficient, or might be in favor of the flag we discussed on Friday for allowing this memoization without making it the default, but I'm 👍 on this as-is.

And apologies if this explanation feels excessive -- seemed worth being thorough considering that this code is used for updating and auditing preserved content, and seemed worth documenting what I looked at, since we're all often a bit uncertain when making changes to important functionality in this gem.

jmartin-sul:
> i have no prediction as to how this would play with ceph issues, either in terms of the possibly increased update throughput, or possibly reduced file read load 🤷

still true, but i don't think we can let that sort of uncertainty keep us from making any changes to our preservation code (even that same uncertainty did contribute to our hesitancy to switch pres cat to sidekiq as soon as we'd have liked).

🎲 we'll just have to take our chances and see if this has any noticeable influence on the onging ceph read issues.

PR #224, opened 2022-08-12
use spec_helper constants for test druid

ndushay:

Let's call it what it is:  OCD.


jcoyne (approved):
😆

PR #226, opened 2022-08-15
storage_repository_spec: update to current rspec conventions

ndushay:

So this spec requires less cognitive load.

Part of #194

### DIFF
--- a/lib/moab/storage_repository.rb
+++ b/lib/moab/storage_repository.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'find' # for object_size
+
 module Moab
   # A class to represent the SDR repository store
   #
--- a/spec/unit_tests/moab/storage_repository_spec.rb
+++ b/spec/unit_tests/moab/storage_repository_spec.rb
@@ -2,48 +2,70 @@
 
 describe Moab::StorageRepository do
   let(:storage_repo) { described_class.new }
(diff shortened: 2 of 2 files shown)


lib/moab/storage_repository.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'find' # for object_size
```
ndushay, line 3: oddly, this isn't necessary unless you run this class's spec on its own;  maybe it is required somewhere in some dependency pulled in (?)

spec/unit_tests/moab/storage_repository_spec.rb
```diff
@@ -66,52 +88,73 @@
...
   describe '#storage_object' do
+    let(:mock_path) { double(Pathname) }
+    let(:mock_storage_obj) { double(Moab::StorageObject) }
+
     before do
-      mock_storage_obj = double(Moab::StorageObject)
       allow(storage_repo).to receive(:find_storage_object).and_return(mock_storage_obj)
-      @mock_path = double(Pathname)
-      allow(mock_storage_obj).to receive(:object_pathname).and_return(@mock_path)
-      allow(@mock_path).to receive(:exist?).and_return(false)
```
ndushay, line 87: die, instance variables, die!

justinlittman (approved):

PR #244, opened 2022-10-18
Remove search_storage_objects.

### LINKED ISSUE #198: remove search_storage_objects method and supporting code (code for multiple moab instances)
`search_storage_objects method` and supporting code was added in PR #162 ;   this is no longer used and no longer needed, since PR sul-dlss/preservation_catalog/pull/1946 was merged.

justinlittman:
Hold pending https://github.com/sul-dlss/preservation_catalog/issues/1891

closes #198

Deprecated


ndushay (changes requested):
There is no need to hold this PR;  the code that uses the method has already been removed from prescat.  

Could you please remove the fixture object "derivatives2" and trappings from the spec code if it's no longer being used?  See PR #162.

justinlittman:
Thanks for the catch @ndushay. Fixed.

Gemfile.lock
```diff
@@ -26,6 +29,12 @@ GEM
     parallel (1.22.1)
     parser (3.1.2.1)
       ast (~> 2.4.1)
+    pry (0.14.1)
+      coderay (~> 1.1)
+      method_source (~> 1.0)
+    pry-byebug (3.10.1)
```
ndushay, line 35: did `Gemfile` change locally for you, Justin, or was the Gemfile.lock just out of date?  I'm ok with pry-byebug in either case;  just want to make sure the code base is consistent with itself.

ndushay (approved):
approving with one more question for consideration

Gemfile.lock
justinlittman: I think the Gemfile.lock was out of date.

PR #261, opened 2023-01-13
Loosen validation of subdirectory names within content directory

### LINKED ISSUE #258: allow subfolders in the content directory to use any name the filesystem allows
It appears that moab-versioning does not allow these names for the top-level folders of a deposited hierarchy:

- data
- metadata
- content
- manifests

I attempted to accession a hierarchy where the top folder is named data:
```
data/
├── content
│   ├── data
│   │   └── level3.txt
│   └── level2.txt
└── level1.txt
```
This all seemed to work correctly except `moab-versioning` threw this error after the files were copied into preservation:
```
 Error: validate-moab : Problem with Moab validation run on preservation-catalog-qa-02.stanford.edu: [{13=>"Version v0001: content directory has forbidden sub-directory name: vnnnn or [\"data\", \"manifests\", \"content\", \"metadata\"]"}] 
```
The Moab looks like it has the correct structure:
```
yy248yb2831/
└── v0001
    ├── data
    │   ├── content
    │   │   └── data
    │   │       ├── content
    │   │       │   ├── data
    │   │       │   │   └── level3.txt
    │   │       │   └── level2.txt
    │   │       └── level1.txt
    │   └── metadata
    │       ├── cocina.json
    │       ├── contentMetadata.xml
    │       └── versionMetadata.xml
    └── manifests
        ├── fileInventoryDifference.xml
        ├── manifestInventory.xml
        ├── signatureCatalog.xml
        ├── versionAdditions.xml
        └── versionInventory.xml
```
The `data` folder that is part of the content is correctly located as a subfolder of the `content` folder and should not interfere with validation.

Sample druid in QA: https://argo-qa.stanford.edu/view/druid:yy248yb2831

**Expected outcome**

The Moab should be considered valid with any directory name in the content folder. There should be no prohibition on content directory names except for those enforced at the filesystem level.

We especially shouldn't prevent people from naming their folders "data".

mjgiarlo:
# Why was this change made? 🤔

Fixes #258

Now that SDR is beginning to allow users to upload content with path information, some paths within deposited items will contain directories named e.g. `data` or `content` or `metadata`. The moab-versioning gem flags these items as having forbidden content subdirectories. This is *not* part of the Moab specification, and we think this was some past attempt to prevent nested moab directory structures. We are now removing this consideration from the gem.

# How was this change tested? 🤨

CI

### DIFF
--- a/lib/moab/storage_object_validator.rb
+++ b/lib/moab/storage_object_validator.rb
@@ -16,7 +16,7 @@ module Moab
     MANIFEST_INVENTORY_PATH = File.join(MANIFESTS_DIR, 'manifestInventory.xml').freeze
     SIGNATURE_CATALOG_PATH = File.join(MANIFESTS_DIR, 'signatureCatalog.xml').freeze
 
-    FORBIDDEN_CONTENT_SUB_DIRS = ([DATA_DIR, MANIFESTS_DIR] + EXPECTED_DATA_SUB_DIRS).freeze
+    VERSION_DIR_PATTERN = /^v\d{4}$/
 
     # error codes
     INCORRECT_DIR_CONTENTS = 0
@@ -32,7 +32,6 @@ module Moab
     NO_FILES_IN_METADATA_DIR = 10
     NO_FILES_IN_CONTENT_DIR = 11
     CONTENT_SUB_DIRS_DETECTED = 12
-    BAD_SUB_DIR_IN_CONTENT_DIR = 13
 
     attr_reader :storage_obj_path
 
@@ -64,8 +63,7 @@ module Moab
           TEST_OBJECT_VERSIONS_NOT_IN_ORDER => 'Should contain only sequential version directories. Current directories: %<addl>s',
           NO_FILES_IN_METADATA_DIR => 'Version %<addl>s: No files present in metadata dir',
           NO_FILES_IN_CONTENT_DIR => 'Version %<addl>s: No files present in content dir',
-          CONTENT_SUB_DIRS_DETECTED => 'Version %<version>s: content directory should only contain files, not directories. Found directory: %<dir>s',
-          BAD_SUB_DIR_IN_CONTENT_DIR => "Version %<addl>s: content directory has forbidden sub-directory name: vnnnn or #{FORBIDDEN_CONTENT_SUB_DIRS}"
+          CONTENT_SUB_DIRS_DETECTED => 'Version %<version>s: content directory should only contain files, not directories. Found directory: %<dir>s'
         }.freeze
     end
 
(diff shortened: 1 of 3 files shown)


lib/moab/storage_object_validator.rb
```diff
@@ -85,7 +83,7 @@ def check_correctly_named_version_dirs
     end
 
     def version_dir_format?(dirname)
-      dirname =~ /^v\d{4}$/
+      /^v\d{4}$/.match?(dirname)
```
mjgiarlo, line 86: We don't use the match object, so don't spend time making one.

lib/moab/storage_object_validator.rb
```diff
@@ -158,11 +156,7 @@ def check_optional_content_dir(version_path, allow_content_subdirs = true)
     end
 
     def contains_forbidden_content_sub_dir?(path)
-      sub_dirs(path).detect { |sub_dir| content_sub_dir_forbidden?(sub_dir) }
-    end
-
-    def content_sub_dir_forbidden?(dirname)
-      FORBIDDEN_CONTENT_SUB_DIRS.include?(dirname) || version_dir_format?(dirname)
+      sub_dirs(path).detect { |sub_dir| version_dir_format?(sub_dir) }
```
mjgiarlo, line 159: Now only vNNNN directories will be forbidden.

spec/unit_tests/stanford/storage_object_validator_spec.rb
```diff
@@ -14,7 +14,7 @@
   describe '#validation_errors' do
     context 'when superclass has validation_errors' do
       it 'returns errors' do
-        expect(error_list.count).to eq(14)
+        expect(error_list.count).to eq(10)
```
mjgiarlo, line 17: Why a difference of four? 

1. data dir
2. content dir
3. manifests dir 
4.  metadata dir

All allowed now.

lib/moab/storage_object_validator.rb
andrewjbtw: Why prohibit those directory names?

mjgiarlo: @andrewjbtw It doesn't seem to be in the Moab "spec" (which is more of a descriptive paper than a normative spec, AFAICT), so... I don't know. Any reason to hang on to this, @andrewjbtw @sul-dlss/infrastructure-team @ndushay ?

ndushay: I think it might be vestigial from our first storage root before ceph having pre-moabs in a diff form.  @julianmorley ?

ndushay (approved):
provisionally approving;  would like Julian's approval.

mjgiarlo:
@ndushay Sounds like Julian doesn't think we need to care much about content subdirectory names. I'll make the change requested by @andrewjbtw above and then unhold the PR, I suppose?

andrewjbtw:
Sounds good to me.

PR #267, opened 2023-01-27
Add performance hacks to FileGroup and FileSignature classes

### LINKED ISSUE #266: SPIKE: Test using Ox instead of nokogiri in content diffing for better performance
Testing on stage has revealed that some objects with many files will time out in the Shelve step (run as a job in DSA). 

See https://app.honeybadger.io/projects/50568/faults/86326783

The timeout occurs when the shelving job in DSA makes a call to a preservation-catalog API endpoint (via the preservation-client) to determine what needs to be shelved.  This call on the preservation-catalog server can take a long time to complete, as it has to parse the contentMetadata XML passed to it by DSA.   It does with nokogiri, which takes a long time for very large blobs of XML.

As a response to this, we are increasing the timeout for preservation-catalog calls: https://github.com/sul-dlss/preservation-client/pull/92

However, even with this increased timeout, some objects on stage continue to timeout.

Some example objects in stage:

druid='druid:pp270vv9568' # 10k files...works
druid='druid:kh844yz4077' # 9k files...doesn't work
druid='druid:gs534db6435' # 25k files... works

Since nokogiri seems to be one of the bottlenecks, this spike would look at the use of nokogiri in the content diffing methods used by the API call that is timing out and see if any performance benefit can be gained by switching to a different XML parsing library (like ox).

Code Archeology:

1. The line in DSA that throws the HB alert is the call to preservation-catalog that times out: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/shelving_service.rb#L48
2. This method in preservation-client calls the preservation-catalog API: https://github.com/sul-dlss/preservation-client/blob/main/lib/preservation/client/objects.rb#L25-L29
3. This method in preservation catalog calls a service wrapper: https://github.com/sul-dlss/preservation_catalog/blob/main/app/controllers/objects_controller.rb#L97-L108
4. The service wrapper then invokes a method in moab-versioning: https://github.com/sul-dlss/preservation_catalog/blob/main/app/services/moab_on_storage/storage_services_wrapper.rb#L15-L20
5. This is the method in moab-versioning invoked:  https://github.com/sul-dlss/moab-versioning/blob/main/lib/stanford/storage_services.rb#L29-L47
6. It then invokes this: https://github.com/sul-dlss/moab-versioning/blob/main/lib/stanford/content_inventory.rb#L20-L30
7. And finally gets here to parse the XML: https://github.com/sul-dlss/moab-versioning/blob/main/lib/stanford/content_inventory.rb#L37-L59

Testing on the rails console on preservation-catalog, starting with this druid on stage: druid:kh844yz4077, and then working your way over to preservation-catalog and running the various steps reveals that the `group_from_cm` method in moab-versioning (step 7 above) is expensive computationally, as it is where the XML parsing occurs.  Previous steps also have a time cost, but I suspect it's the XML parsing that is the most expensive.

mjgiarlo:

Fixes #266

This commit changes the `FileGroup`, `FileInventoryDifference`, `FileSignature`, and `ContentInventory` classes in order to reduce performance bottlenecks when diffing moabs with voluminous content metadata. (See #266 for a breakdown of the profiling results and methodology.)

Generally, the changes were intended to avoid unnecessary mutations, to make more judicious use of iterators (more important when dealing with extremely large collections!), and to cut down on the sheer number of method invocations within operations being called millions of times for a large test object. Each of these methods were targeted based on [profiling analysis](https://github.com/sul-dlss/moab-versioning/issues/266#issuecomment-1406812866).

Specific changes include:

* `FileGroup#path_hash_subset`
  * Prefer filtering the `FileGroup#signature_hash` earlier in the process to avoid iterating over irrelevant entries
  * Prefer using the result of `Enumerable#map` over using `Enumerable#each` to incrementally mutate a hash
* `FileGroup#add_file_instance`
  * Sniff if `Hash#[]` returns `nil` instead of using `Hash#key?`,
* `FileInventoryDifference#compare`
  * Use the result of `Enumerable#map` instead of using `Enumerable#each` plus mutation via shovel operator
* `FileSignature`
  * Replace `n.nil? ? '' : n.to_s` with `n.to_s` in attr `on_save` procs, since `nil.to_s` returns `''`
* `FileSignature#checksums`
  * Prefer `Object#tap` and injecting checksums if present over calling `Hash#reject` and asking every value if it's either `nil?` or `empty?`
* `FileSignature#eql?`
  * Rescue `NoMethodError` instead of calling `respond_to?` twice for every call

CI + stage

### DIFF
--- a/lib/moab/file_group.rb
+++ b/lib/moab/file_group.rb
@@ -97,16 +97,23 @@ module Moab
 
     # @api internal
     # @param signature_subset [Array<FileSignature>] The signatures used to select the entries to return
-    # @return [Hash<String,FileSignature>] A pathname,signature hash containing a subset of the filenames in this file group
+    # @return [Hash<String,FileSignature>] A pathname,signature hash containing a subset of the filenames in this file
+    #                                      group, e.g., {"intro-1.jpg"=>#<Moab::FileSignature>, ...}
     def path_hash_subset(signature_subset)
-      path_hash = {}
-      signature_subset.each do |signature|
-        manifestation = signature_hash[signature]
-        manifestation.instances.each do |instance|
-          path_hash[instance.path] = signature
-        end
+      # the structure of the `signature_hash` attr is documented above
+      signature_hash
+        .filter_map do |signature, manifestation|
+        # filters out signatures not in the provided subset
+        next unless signature_subset.include?(signature)
+
+        # for each instance in the manifestation, return an array of its path and the signature from the above block
+        manifestation.instances.map { |instance| [instance.path, signature] }
       end
-      path_hash
+        # the nested map operations above return e.g.: [[["intro-1.jpg",
+        # #<Moab::FileSignature>],...]] which needs to be flattened one time to
+        # convert back into a hash
+        .flatten(1)
+        .to_h
     end
 
     # @param manifestiation_array [Array<FileManifestation>] The collection of {FileManifestation} objects
@@ -133,12 +140,11 @@ module Moab
     # @return [void] Add a single {FileSignature},{FileInstance} key/value pair to this group.
     #   Data is actually stored in the {#signature_hash}
     def add_file_instance(signature, instance)
-      if signature_hash.key?(signature)
-        manifestation = signature_hash[signature]
-      else
-        manifestation = FileManifestation.new
-        manifestation.signature = signature
-        signature_hash[signature] = manifestation
+      manifestation = signature_hash[signature] || begin
+        FileManifestation.new.tap do |file_manifestation|
+          file_manifestation.signature = signature
+          signature_hash[signature] = file_manifestation
+        end
       end
       manifestation.instances << instance
     end
--- a/lib/moab/file_signature.rb
+++ b/lib/moab/file_signature.rb
@@ -45,20 +45,20 @@ module Moab
     tag 'fileSignature'
 
     # @attribute
-    # @return [Integer] The size of the file in bytes
+    # @return [String] The size of the file in bytes (can be empty)
     attribute :size, Integer, on_save: proc { |n| n.to_s }
 
     # @attribute
-    # @return [String] The MD5 checksum value of the file
-    attribute :md5, String, on_save: proc { |n| n.nil? ? '' : n.to_s }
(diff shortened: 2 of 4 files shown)


ndushay (approved):
LGTM.  How good are the tests?  Were you able to break them?

lib/moab/file_group.rb
```diff
@@ -99,14 +99,10 @@ def path_list
...
-      path_hash = {}
-      signature_subset.each do |signature|
-        manifestation = signature_hash[signature]
-        manifestation.instances.each do |instance|
-          path_hash[instance.path] = signature
-        end
-      end
-      path_hash
+      signature_hash
+        .select { |signature, _manifestation| signature_subset.include?(signature) }
```
mjgiarlo, line 103: Consider how to do this with only one iteration over the full data set.

lib/moab/file_group.rb
```diff
@@ -99,14 +99,10 @@ def path_list
...
     def path_hash_subset(signature_subset)
-      path_hash = {}
-      signature_subset.each do |signature|
-        manifestation = signature_hash[signature]
-        manifestation.instances.each do |instance|
-          path_hash[instance.path] = signature
-        end
-      end
-      path_hash
+      signature_hash
```
mjgiarlo, line 102: Consider giving examples of the data.

lib/moab/file_signature.rb
```diff
@@ -50,15 +50,15 @@ class FileSignature < Serializer::Serializable
 
     # @attribute
     # @return [String] The MD5 checksum value of the file
-    attribute :md5, String, on_save: proc { |n| n.nil? ? '' : n.to_s }
+    attribute :md5, String, on_save: proc { |n| n.to_s }
```
mjgiarlo, line 53: Add to comment: `can be empty string`

mjgiarlo:
@ndushay 💬 
> LGTM. How good are the tests? Were you able to break them?

I didn't try to break the tests. I'm trying to improve moab-versioning performance using internal method refactorings that preserve their current behavior, so I intentionally did not touch the tests in this work (other than running them and making sure they pass!). I share your worries about how many of the tests in moab-versioning are good or legit tests, but I guess I see this as separate from the effort to improve performance.

